PR #2, opened 2024-05-02
Working PoC for Additional Marker Attribute

Tornhoof:
Prototype to support Additional Marker Attribute

### DIFF
--- a/src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
+++ b/src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
@@ -1,0 +1,27 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using Microsoft.CodeAnalysis;
+
+namespace TypeShape.SourceGenerator.Helpers;
+
+internal static partial class RoslynHelpers
+{
+    public static bool TryGetAdditionalMarkerAttributeName(this IncrementalGeneratorInitializationContext context,
+        [MaybeNullWhen(false)] out string attributeName)
+    {
+        string? configOption = null;
+        var valueProvider = context.AnalyzerConfigOptionsProvider
+            .Select((options, _) =>
+            {
+                if (options.GlobalOptions
+                    .TryGetValue("build_property.TypeShape_SourceGenerator_AdditionalMarkerAttributeName",
+                        out var markerAttribute))
+                    configOption = markerAttribute;
+
+                return 0;
+            });
+        context.RegisterSourceOutput(valueProvider, (_, __) => { });
+
+        attributeName = configOption;
+        return configOption is not null;
+    }
+}
--- a/src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
+++ b/src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
@@ -1,27 +1,199 @@
-﻿using System.Diagnostics.CodeAnalysis;
+﻿using System.Collections.Immutable;
+using System.Diagnostics;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
 
 namespace TypeShape.SourceGenerator.Helpers;
 
 internal static partial class RoslynHelpers
 {
-    public static bool TryGetAdditionalMarkerAttributeName(this IncrementalGeneratorInitializationContext context,
-        [MaybeNullWhen(false)] out string attributeName)
+    public static void RegisterAdditionalMarkerAttributeName(this IncrementalGeneratorInitializationContext context,
+        MarkerAttributeHolder holder)
     {
-        string? configOption = null;
         var valueProvider = context.AnalyzerConfigOptionsProvider
             .Select((options, _) =>
             {
-                if (options.GlobalOptions
-                    .TryGetValue("build_property.TypeShape_SourceGenerator_AdditionalMarkerAttributeName",
+                if (options.GlobalOptions.TryGetValue(
+                        "build_property.TypeShape_SourceGenerator_AdditionalMarkerAttributeName",
                         out var markerAttribute))
-                    configOption = markerAttribute;
+                {
+                    NameSyntax attributeNameSyntax = SyntaxFactory.ParseName(markerAttribute);
+                    string attributeName = GetTypeName(attributeNameSyntax, out int attributeArity);
+                    ImmutableArray<string> attributeNamespace = GetNamespaceTokens(attributeNameSyntax);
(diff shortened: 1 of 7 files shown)


src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
```diff
@@ -0,0 +1,199 @@
+﻿using System.Collections.Immutable;
+using System.Diagnostics;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+
```
eiriktsarpalis, line 7:
```suggestion

```

src/TypeShape.SourceGenerator/TypeShapeIncrementalGenerator.cs
```diff
@@ -12,7 +12,11 @@ public sealed class TypeShapeIncrementalGenerator : IIncrementalGenerator
     public void Initialize(IncrementalGeneratorInitializationContext context)
     {
 #if LAUNCH_DEBUGGER
-        System.Diagnostics.Debugger.Launch();
+        if (!System.Diagnostics.Debugger.IsAttached)
```
eiriktsarpalis, line 15: I had removed that guard initially because I found it didn't make any difference in practice. Did I miss something?

Tornhoof: I don't think so, I had some weird issues with debugging, so I copied some boilerplate from another srcgen in there, but apparently never reverted that change (probably because it's in #if). I'll remove that.

src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
```diff
@@ -0,0 +1,196 @@
...
+                    string? attributeNameMinusSuffix = attributeName.EndsWith("Attribute", StringComparison.Ordinal)
+                        ? attributeName[..^"Attribute".Length]
+                        : null;
+
+                    holder.Populate(attributeName, attributeNameMinusSuffix, attributeNamespace, attributeArity);
+                }
+
+                return 0;
+            });
+        context.RegisterSourceOutput(valueProvider, (_, _) => { }); // this does not produce anything, but is required for the provider to be executed?
```
eiriktsarpalis, line 33: That's surprising, my understanding is that this can only happen if the incremental value somehow doesn't flow to another `RegisterSourceOutput` call.

src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
```diff
@@ -0,0 +1,196 @@
...
+                        out var markerAttribute))
+                {
+                    NameSyntax attributeNameSyntax = SyntaxFactory.ParseName(markerAttribute);
+                    string attributeName = GetTypeName(attributeNameSyntax, out int attributeArity);
+                    ImmutableArray<string> attributeNamespace = GetNamespaceTokens(attributeNameSyntax);
+                    string? attributeNameMinusSuffix = attributeName.EndsWith("Attribute", StringComparison.Ordinal)
+                        ? attributeName[..^"Attribute".Length]
+                        : null;
+
+                    holder.Populate(attributeName, attributeNameMinusSuffix, attributeNamespace, attributeArity);
```
eiriktsarpalis, line 28: Why does this need to use global mutable state?

src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
```diff
@@ -0,0 +1,196 @@
...
+
+                return 0;
+            });
+        context.RegisterSourceOutput(valueProvider, (_, _) => { }); // this does not produce anything, but is required for the provider to be executed?
+
+        static string GetTypeName(NameSyntax nameSyntax, out int genericTypeArity)
+        {
+            while (true)
+            {
+                switch (nameSyntax)
```
eiriktsarpalis, line 39: This appears to be replicating logic from the other helper?

src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
```diff
@@ -0,0 +1,196 @@
...
+
+internal static partial class RoslynHelpers
+{
+    public static void RegisterAdditionalMarkerAttributeName(this IncrementalGeneratorInitializationContext context,
+        MarkerAttributeHolder holder)
+    {
+        var valueProvider = context.AnalyzerConfigOptionsProvider
+            .Select((options, _) =>
+            {
+                if (options.GlobalOptions.TryGetValue(
```
eiriktsarpalis, line 17: Assuming this gains traction, it might be necessary to support multiple attributes. Would that be possible, e.g. using multiple entries or comma-separated values?

src/TypeShape.SourceGenerator/Helpers/RoslynHelpers.AdditionalMarkerAttribute.cs
```diff
@@ -0,0 +1,196 @@
...
+                        "build_property.TypeShape_SourceGenerator_AdditionalMarkerAttributeName",
+                        out var markerAttribute))
+                {
+                    NameSyntax attributeNameSyntax = SyntaxFactory.ParseName(markerAttribute);
+                    string attributeName = GetTypeName(attributeNameSyntax, out int attributeArity);
```
eiriktsarpalis, line 22: We should probably reject any types that are generic (i.e. that have arity > 0)

src/TypeShape.SourceGenerator/TypeShapeIncrementalGenerator.cs
```diff
@@ -22,20 +22,30 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
...
-            .Select((tuple, cancellationToken) => 
+            .Select((tuple, cancellationToken) =>
                 Parser.ParseFromGenerateShapeOfTAttributes(
                     context: tuple.Left,
-                    knownSymbols: tuple.Right, 
+                    knownSymbols: tuple.Right,
                     cancellationToken));
 
+
+        // I kinda expect this to break the whole incremental pipeline
```
eiriktsarpalis, line 32: My expectation here would be that this should be fused with the incremental values defined in line 39. The `ForTypeWithAttributeDeclaration` helper could be augmented with an optional `GlobalOptions` key so that the search is restricted to one pass (and should serve to collapse duplicates in cases where a type is annotated with both `GenerateShape` and the user-defined attribute). 

Such a change would necessitate lifting the extension method from the syntax provider to the `IncrementalGeneratorInitializationContext` itself so that the helper can access the `AnalyzerConfigOptionsProvider`.

PR #4, opened 2024-05-03
Fix Size Check

Tornhoof:
In the strategies in the same method, both the size is checked and elementSize is passed to the element functor, but not in the main functor. This fixes that and adds a test which asserts without the fix.


tests/TypeShape.Tests/BugTests.cs
```diff
@@ -0,0 +1,26 @@
+﻿using TypeShape.Applications.RandomGenerator;
+using Xunit;
+
+namespace TypeShape.Tests
+{
+    public partial class BugTests
+    {
+        [Fact]
```
eiriktsarpalis, line 8: Could you move the test to the existing random generator test suite? Thanks

Tornhoof: Moved it over, need to exclude it from the JsonSchema tests, it can't handle the recursive type def apparently. It dies with StackOverflow Exception,
```csharp
 	TypeShape.dll!TypeShape.ReflectionProvider.ReflectionHelpers.GetNullabilityInfo(System.Reflection.ICustomAttributeProvider memberInfo = {System.Reflection.RuntimePropertyInfo}) Line 112	C#
 	TypeShape.dll!TypeShape.ReflectionProvider.ReflectionHelpers.ResolveNullableAnnotation(System.Reflection.MemberInfo memberInfo = {System.Reflection.RuntimePropertyInfo}, out bool isGetterNonNullable = true, out bool isSetterNonNullable = false) Line 32	C#
 	TypeShape.dll!TypeShape.ReflectionProvider.MemberInitializerShapeInfo.MemberInitializerShapeInfo(System.Reflection.MemberInfo memberInfo = {System.Reflection.RuntimePropertyInfo}, string logicalName = null) Line 109	C#
>	TypeShape.dll!TypeShape.ReflectionProvider.ReflectionTypeShape<TypeShape.Tests.ClassWithMultipleSelfReferences>.GetConstructors.AnonymousMethod__16_1((System.Reflection.MemberInfo MemberInfo, System.Reflection.ICustomAttributeProvider AttributeProvider, string LogicalName, int Order, bool IncludeNonPublic) m = ({System.Reflection.RuntimePropertyInfo}, {System.Reflection.RuntimePropertyInfo}, null, 0, false)) Line 49	C#

```

eiriktsarpalis: Interesting. The standard test suite does include recursive types already so I suppose that's a different bug?

Tornhoof: I modified the test type to prevent the stackoverflow exception.

PR #14, opened 2024-10-24
Add a reference to MessagePack implementation

AArnott:
This pull request includes updates to the `README.md` file, enhancements to the `SpanDictionary` class, and improvements to the equality members of the `BaseClass` and `DerivedClass` in the test types.

### Documentation Updates:
* [`README.md`](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R101-R106): Added a new section listing known libraries based on TypeShape, including `Nerdbank.MessagePack`.

### Code Enhancements:
* [`src/TypeShape.Examples/Utilities/SpanDictionary.cs`](diffhunk://#diff-8be3382a3e32b1cd3cd00276953a089c5566db5b9a0b9bea487ce21b08ffba1eR141-R144): Added a new overload for the `ToSpanDictionary` method to support mapping with a value selector.

### Test Improvements:
* [`tests/TypeShape.Tests/TestTypes.cs`](diffhunk://#diff-0083fee13231a2555b1cd0cec692f228418a655187e39e28414bf62261ef48a9L595-R615): Implemented `IEquatable<T>` for `BaseClass` and `DerivedClass`, and added appropriate `Equals` and `GetHashCode` methods.


AArnott:
Incidentally, upgrading the baseline from MessagePack v2 to v3 produces this:

| Method                | Mean     | Error   | StdDev  | Ratio | Allocated | Alloc Ratio |
|---------------------- |---------:|--------:|--------:|------:|----------:|------------:|
| Deserialize_TypeShape | 186.8 ns | 1.21 ns | 1.13 ns |  1.54 |      96 B |        3.00 |
| Deserialize_Library   | 120.9 ns | 1.06 ns | 0.94 ns |  1.00 |      32 B |        1.00 |

| Method              | Mean     | Error   | StdDev  | Ratio | Allocated | Alloc Ratio |
|-------------------- |---------:|--------:|--------:|------:|----------:|------------:|
| Serialize_TypeShape | 149.5 ns | 0.87 ns | 0.81 ns |  1.06 |         - |          NA |
| Serialize_Library   | 140.8 ns | 0.71 ns | 0.66 ns |  1.00 |         - |          NA |

This improves serialization slightly, making TypeShape fair slightly worse by comparison.
Deserialization improves significantly, making TypeShape look quite a bit worse. Still interested in whether that can be fixed.

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
```diff
@@ -0,0 +1,282 @@
...
+    public static T? Deserialize<T>(ReadOnlySequence<byte> sequence)
+        where T : IShapeable<T>
+    {
+        MessagePackReader reader = new(sequence);
+        return SerializerCache<T>.Deserializer(ref reader);
+    }
+
+    private delegate T? Reader<T>(ref MessagePackReader reader);
+
+    private delegate void Writer<T>(ref MessagePackWriter writer, T value);
```
eiriktsarpalis, line 44: Couldn't the two delegates be squished into a common `MessagePackConverter<T>` or `MessagePackFormatter<T>` type? Or are you intentionally looking to separate the two?

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
```diff
@@ -0,0 +1,282 @@
...
+        internal static Writer<T?> Serializer => serializer ??= SerializeVisitor.GetWriter(T.GetShape());
+
+        internal static Reader<T?> Deserializer => deserializer ??= DeserializeVisitor.GetReader(T.GetShape());
+    }
+
+    private sealed class SerializeVisitor : TypeShapeVisitor
+    {
+        internal static readonly SerializeVisitor Instance = new();
+
+        private static readonly TypeDictionary formatters = new()
```
eiriktsarpalis, line 62: The dictionary isn't thread safe, as you pointed out in the issue that you created.

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
```diff
@@ -0,0 +1,282 @@
...
+                    writer.WriteNil();
+                    return;
+                }
+
+                writer.WriteMapHeader(properties.Count);
+
+                foreach (var property in properties)
+                {
+                    writer.WriteRaw(property.Key.Span);
+                    property.Value(ref writer, value);
```
eiriktsarpalis, line 100: If `T` is a struct this would force copying of `value` every time a property is being serialized. Consider using a delegate that accepts it by reference.

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
```diff
@@ -0,0 +1,282 @@
...
+        internal static Writer<T?> GetWriter<T>(ITypeShape<T> typeShape)
+        {
+            return formatters.GetOrAdd<Writer<T?>>(typeShape, Instance, box => (ref MessagePackWriter writer, T? value) => box.Result(ref writer, value));
+        }
+
+        private SerializeVisitor() { }
+
+        public override object? VisitObject<T>(IObjectTypeShape<T> objectShape, object? state = null)
+        {
+            Dictionary<ReadOnlyMemory<byte>, Writer<T>> properties = new();
```
eiriktsarpalis, line 77: Given that you're not looking this dictionary up, it might make sense to change this to `(byte[], Writer<T>)[]` for faster iteration times when serializing.

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
```diff
@@ -0,0 +1,282 @@
...
+            });
+        }
+
+        public override object? VisitProperty<TDeclaringType, TPropertyType>(IPropertyShape<TDeclaringType, TPropertyType> propertyShape, object? state = null)
+        {
+            var getter = propertyShape.GetGetter();
+
+            switch (propertyShape.PropertyType.Kind)
+            {
+                case TypeShapeKind.Object:
```
eiriktsarpalis, line 111: I'm guessing that this switch statement wouldn't be necessary assuming that you're returning a writer for all shape kinds.

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
```diff
@@ -0,0 +1,282 @@
...
+        private DeserializeVisitor() { }
+
+        private delegate void PropertySetter<T>(ref MessagePackReader reader, ref T container);
+
+        public override object? VisitObject<T>(IObjectTypeShape<T> objectShape, object? state = null)
+        {
+            SpanDictionary<byte, PropertySetter<T>> properties = objectShape.GetProperties()
+                .Where(p => p.HasSetter)
+                .ToSpanDictionary(
+                    p => Encoding.UTF8.GetBytes(p.Name),
```
eiriktsarpalis, line 152: Does message pack support any kind of escaping or case insensitivity in its property names that would make ordinal UTF-8 comparison unsuitable?

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
AArnott: For what it's worth, messagepack has users where they can only serialize or deserialize a type, not both. So it made sense to me to write this so that they were independently generated.

AArnott: That said, if I end up implementing this as an `IMessagePackFormatter<T>`, then yes I'll be implementing both directions as one unit.

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
AArnott: Confirmed that once I change this to `ref`, the `ref` ultimately bleeds out to my top-level `MsgPackSerializer.Serialize` method, which now has to take `ref` instead of the more implicit-friendly `scoped in` parameter.

AArnott:
> Serialization is somehow faster than the Ref.Emit specialized code produced by the MessagePack library. I don't know how that's possible.

I suspect it's because in this version, we call `MessagePackWriter.Write(string)` directly rather than including a level of indirection that allows for some other serializer to hijack that operation at runtime to serialize something else (e.g. so that a string is only serialized once).

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
eiriktsarpalis: > What do you think? Should/can this be changed to `scoped in` rather than `ref`?

There are circumstances where passing a mutable struct with `in` prompts the codegen to create defensive copies, although I'm not sure if this applies to generic methods. Using ref should result in simpler codegen because it has simpler semantics.

> Confirmed that once I change this to `ref`, the `ref` ultimately bleeds out to my top-level

It's a problem if it leaks to user-facing API. What I usually do to address the issue is author [property-specific delegates/converters](https://github.com/eiriktsarpalis/typeshape-csharp/blob/c8c2c397beb472e716b447bb460dea626811ba18/src/TypeShape.Examples/JsonSerializer/Converters/JsonPropertyConverter.cs#L16-L17) since their signature tends to differ especially for the case of deserialization/binding.

AArnott: I could do that I guess. It means that the Serialize path makes 1 copy of the struct where it could have made 0 though.

Use of generics itself doesn't lead to the defensive copies. It's the C# compiler that does it when invoking a method/property accessor on a struct that is stored in a read-only field but the method isn't labeled `readonly`. 
Consider [this example](https://sharplab.io/#v2:EYLgHgbALANALiAlgGwD4AEBMBGAsAKAKwAIBlAewFsBTAbwOMeMQDs5iAxc8gCleIDCxAG4BDZAFdqASgZN6+JkvQB2MlWoBxanDjUATjzGSZSgNxzGAX2KXbipgBNqyagHNRe4gBUAauKkfbV0DAB5vAXI2UVYDGB9/EwA+PhYfSOjY/WIAYyi4GJYDaQsHRjt0bAggnT19UIF41jgk9RpguuIAXmJUwVzpbtacgDoABX1yAAcDOABPUqsCIgBmYgBnOH0JHPYhBWU15uIJ6dm5oeIAVkWCIA=) where I add some use of generics to make things interesting, but it's basically what I'm proposing for your property getter delegate in that it uses `in` instead of `ref`. Here, because the property getter is manually defined, and lacks a `readonly` modifier, a defensive copy is made before invoking `get_Property`:

```
IL_0000: ldarg.1
IL_0001: ldobj C
IL_0006: stloc.0
IL_0007: ldloca.s 0
IL_0009: call instance int32 C::get_Property()
IL_000e: ret
```

But just add `readonly` to the property, and the same invocation reduces to:

```
IL_0000: ldarg.1
IL_0001: call instance int32 C::get_Property()
IL_0006: ret
```

So we have a choice: always do 1 copy of the struct as part of serialization to avoid any further copies. Or shoot for 0 copies of the struct by using `in` instead of `ref`, but acknowledge you may have *n* copies in reality if you're serializing a struct with custom property getters that don't use `readonly` on them.
Most custom property getter authors probably *don't* bother to add `readonly` to them. So shooting for exactly 1 (by using `ref`) will be faster on average. But if we don't mind shooting for 0 and pointing the blame at others to fix their own code if they complain about excess copies, then using `in` is better.

eiriktsarpalis: It's probably not the end of the world if only one copy is being made per node (the STJ converters also serialize by value) but if we really wanted to we could make the core delegate/formatter also accept values by ref. In that case we'd be dealing with only one copy for the root-level value in the outer method that resolves the `IShapeable<T>`.

AArnott: In my latest design where I generate `IMessagePackFormatter<T>` (the same currency that MessagePack-CSharp uses), it's a non-issue anyway because we copy structs to serialize them (sadly, but that's the interface). I _do_ use `ref` for internally passing the values around though.

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
```diff
@@ -0,0 +1,345 @@
...
+        public override object? VisitConstructor<TDeclaringType, TArgumentState>(IConstructorShape<TDeclaringType, TArgumentState> constructorShape, object? state = null)
+        {
+            return constructorShape.GetDefaultConstructor();
+        }
+    }
+
+    private delegate T? FormatDeserialize<T>(ref MessagePackReader reader, MessagePackSerializerOptions options);
+    private delegate void FormatSerialize<T>(ref MessagePackWriter writer, T? value, MessagePackSerializerOptions options);
+
+    private sealed class Formatter<T>(FormatDeserialize<T> deserialize, FormatSerialize<T> serialize) : IMessagePackFormatter<T?>
```
eiriktsarpalis, line 340: I can see why you went for a delegate adapter, but it does result in two layers of indirection for every call, which adds up as the object graph is being traversed.

AArnott: Fixed.
Can you review all the places where `formatters.GetOrAdd` appears in my code and tell me if I'm doing it right? I don't understand why this pattern is required, so I'm not sure...

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
```diff
@@ -232,18 +232,20 @@ private sealed class FormatterVisitor : TypeShapeVisitor
...
-            { typeof(int), new Formatter<int>((ref MessagePackReader reader, MessagePackSerializerOptions options) => reader.ReadInt32(), (ref MessagePackWriter writer, int value, MessagePackSerializerOptions options) => writer.Write(value)) },
+            { typeof(string), new PrimitiveFormatter<string>((ref MessagePackReader reader) => reader.ReadString(), (ref MessagePackWriter writer, string? value) => writer.Write(value)) },
+            { typeof(int), new PrimitiveFormatter<int>((ref MessagePackReader reader) => reader.ReadInt32(), (ref MessagePackWriter writer, int value) => writer.Write(value)) },
         };
 
         internal IMessagePackFormatter<T?> GetFormatter<T>(ITypeShape<T> typeShape)
         {
-            return formatters.GetOrAdd<IMessagePackFormatter<T?>>(typeShape, this, box => new Formatter<T?>(box.Result.Deserialize, box.Result.Serialize));
+            return formatters.GetOrAdd<IMessagePackFormatter<T?>>(typeShape, this, box => box.Result switch
```
eiriktsarpalis, line 241: The delayed value factory is a late binding mechanism used when a recursive type is encountered. The idea is that it's creating a wrapper for `box.Result` which will _eventually_ contain a reference to the final computed value once building has completed. 

For this to work, evaluating `box.Result` needs to be delayed until an actual serialization operation is run. Here's how it's done for CBOR:

https://github.com/eiriktsarpalis/typeshape-csharp/blob/c8c2c397beb472e716b447bb460dea626811ba18/src/TypeShape.Examples/CborSerializer/Converters/DelayedCborConverter.cs#L6-L13

And here's how the object cloner does it:

https://github.com/eiriktsarpalis/typeshape-csharp/blob/main/src/TypeShape.Examples/Cloner/Cloner.cs#L62

src/TypeShape.Examples/MsgPackSerializer/MsgPackSerializer.cs
eiriktsarpalis: I was suggesting that you could make `Formatter<T>` an abstract class and have individual types and shapes inherit from that. The downside is that you need to factor those into individual classes as opposed to inlining lambda expressions, but on the other hand splitting your code into a folder of Formatters and then a keeping a separate builder that folds them together should make things more manageable.

AArnott: So far, I'm not sure what I'd do with more than the two implementations I have.
Maybe I'll need more as I get to supporting non-default constructors, etc. though.

AArnott:
@neuecc this should interest you.

tests/TypeShape.Tests/TestTypes.cs
```diff
@@ -592,15 +592,27 @@ public StructWithDefaultCtor()
 }
 
 [GenerateShape]
-public partial class BaseClass
+public partial class BaseClass : IEquatable<BaseClass>
```
eiriktsarpalis, line 595: Would it make sense to factor the IEquatable implementation to a separate set of types so that we coverage for both cases?

AArnott: I added this because I had been testing round-tripping. I'm not sure what case we're *not* covering when this interface isn't present. 
But I _did_ just discover `IsEquatable` on test cases, which probably was `false` for this type up to this point and why no one else hit the fact that this class didn't know how to do a by-value compare of itself. 
I can remove this change if you'd like. I don't need it any more.

PR #22, opened 2024-10-25
Use strong naming for netstandard2.0 assemblies

eiriktsarpalis:
cc @AArnott 

I'm only signing the Roslyn/netstandard2.0 assemblies and skipping it for the primary assemblies that are currently `net8.0` only.


AArnott:
> I'm only signing the Roslyn/netstandard2.0 assemblies and skipping it for the primary assemblies that are currently net8.0 only.

That's fine.
I never understood though the reason folks (not just you) avoid strong naming as if it presented a problem.

PR #23, opened 2024-10-25
Apply version stamps with Nerdbank.GitVersioning

AArnott:
This replaces MinVer with [Nerdbank.GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning/).

It also applies the policy that `TypeShape.Roslyn` should be precisely versioned, which is one of the two requirements for safe consumption from roslyn analyzers and source generators.

BTW, an optional and useful tool to use in combination with this is [`nbgv`](https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/nbgv-cli.md), which can automate releases, tagging, etc.


.github/workflows/build.yml
```diff
@@ -15,6 +15,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0 # avoid shallow clone so nbgv can do its work.
```
eiriktsarpalis, line 19: Does it need to clone the entire history to work? Can it be truncated to a reasonable upper bound perhaps?

Directory.Packages.props
```diff
@@ -28,4 +27,7 @@
     <PackageVersion Include="JsonSchema.Net" Version="7.2.3" />
     <PackageVersion Include="FSharp.Core" Version="8.0.401" />
   </ItemGroup>
-</Project>
\ No newline at end of file
+  <ItemGroup>
+    <GlobalPackageReference Include="Nerdbank.GitVersioning" Version="3.6.146" />
```
eiriktsarpalis, line 31: I would probably only scope this to projects under the src/ folder

.github/workflows/build.yml
AArnott: It only needs to be high enough to reach the last change to version.json. But in practice, that's a hard number to estimate, and unless you're in a repo like VS, shallow clones don't really save significant time on clone in my experience.

Directory.Packages.props
AArnott: That's an option. But why bother to add nuance?

src/TypeShape.Roslyn/version.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/main/src/NerdBank.GitVersioning/version.schema.json",
+  "inherit": true,
+  "assemblyVersion": {
```
eiriktsarpalis, line 4: Presumably this is refining the assembly version for `TypeShape.Roslyn` specifically? What is the default assembly version for the others?

AArnott: The default is to take major.minor from the `version` property for the assembly version.

PR #25, opened 2024-10-25
Add a strong name to the assemblies

AArnott:
This is one of two requirements for TypeShape.Roslyn to be safe for consumption by roslyn analyzers.


AArnott:
Redundant (and takes a different approach) with #22

PR #31, opened 2024-10-26
Stop writing emitted files to disk

AArnott:
On the second computer I've cloned this repo down to, the build fails because one of the files the compiler tries to write has a file length that's too long:
'D:\source\typeshape-csharp\tests\TypeShape.Tests\obj\Debug\net8.0\generated\TypeShape.SourceGenerator\TypeShape.SourceGenerator.TypeShapeIncrementalGenerator\TypeShape.Tests.SourceGenProvider.ValueTuple_Int32_Int32_Int32_Int32_Int32_Int32_Int32_Int32_Int32.g.cs' is 261 characters long

It's odd that this path is shorter than my first computer, which has no problem with builds (although many tools and features of Windows are still unable to process this file or even delete it).

I suggest in the interest in making this repo more consumable by more people and machines, that we turn off generated file emission on disk. Folks can always look at the generated code through Solution Explorer:
image

Note however that this code will be generated by the _first_ version of the generator that VS could load for that process. Any changes made to the source generator after that won't be reflected in the generated code accessible in Solution Explorer until those changes are built and VS is restarted. So if your "inner loop" involves making changes and reviewing them, turning this property back on (but not checking it in, preferably) would perhaps be the best path.

### DIFF
--- a/tests/TypeShape.Tests/TypeShape.Tests.csproj
+++ b/tests/TypeShape.Tests/TypeShape.Tests.csproj
@@ -6,7 +6,6 @@
     <Nullable>enable</Nullable>
     <WarningsAsErrors>Nullable</WarningsAsErrors>
     <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
-    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
     <DefineConstants>IS_TEST_PROJECT</DefineConstants>
   </PropertyGroup>
 


eiriktsarpalis:
Sure we can turn it off, although my recommendation is to always turn on long path support when developing on Windows.

AArnott:
I did turn it on, before filing this. It didn't help. Although... was I supposed to reboot first?

eiriktsarpalis:
> Although... was I supposed to reboot first?

According to [this article](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later), probably yes.

eiriktsarpalis:
There's also a setting in git that needs to be set separately:

https://github.com/dotnet/runtime/blob/main/docs/workflow/requirements/windows-requirements.md#enable-long-paths

PR #41, opened 2024-11-01
Offer test cases as a nuget package

### LINKED ISSUE #34: Offer nuget package with test cases
Your examples' tests have a massive battery of test cases, which is awesome and useful for serialization libraries built on TypeShape outside your repo. Can you offer a separate nuget package that just contains your test cases for consumption by our test projects?

AArnott:
This moves the fairly exhaustive battery of test cases that all the examples use to a `TypeShape.TestCases` nuget package, where they may be used by other repos to test their TypeShape-based libraries.

Closes #34


AArnott:
Let's not merge this until I verify that I can consume it effectively from externally.

For example:
- I don't have any `IProviderUnderTest` implementations shared in the new package.

AArnott:
You already have an inconsistent package ID format:
- `typeshape-csharp`
- `TypeShape.Examples`
- `TypeShape.Roslyn`

I went with `TypeShape.TestCases` rather than `typeshape-testcases`.

src/TypeShape.TestCases/README.md
```diff
@@ -0,0 +1,14 @@
+# TypeShape consumer test cases
```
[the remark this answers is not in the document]
AArnott: Agreed. I'm modifying the doc to suggest they just use a static provider here. 
_Removing_ the `IProviderUnderTests` here though would presumably be a lot more work.

src/TypeShape.TestCases/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+This package is intended for consumption by your unit test project. It presents a bunch of types that help to exhaustively test your TypeShape visitors.
+
+```cs
+private static readonly IProviderUnderTest Provider = SourceGenProviderUnderTest.Default; // or RefectionProviderUnderTest.Default;
+
+[Theory]
+[MemberData(nameof(TestTypes.GetTestCases), MemberType = typeof(TestTypes))]
+public void Roundtrip_Value<T>(TestCase<T> testCase)
```
eiriktsarpalis, line 10: Is it possible to use the `TestCase<T, TProvider>` as the theory parameter? I suspect it won't work.

AArnott: It apparently _does_.

eiriktsarpalis: Oh interesting. Nice.

AArnott:
> Let's not merge this until I verify that I can consume it effectively from externally.

It works. 
At least, tests are running. But my library fails 73 tests (and passes 247).

src/TypeShape.TestCases/README.md
AArnott: This compiles and runs tests:

```cs
[Theory]
[MemberData(nameof(TestTypes.GetTestCases), MemberType = typeof(TestTypes))]
public void Roundtrip_Value<T, TProvider>(TestCase<T, TProvider> testCase)
	where TProvider : IShapeable<T>
{
	ITypeShape<T> shape = TProvider.GetShape();
```

AArnott: The only change this would make to the package I'm building here is to its README file. I could change it to prescribe these as the recommended pattern, but as the user can already get an `ITypeShape<T>` with the prior/current pattern, is there a benefit that I'm missing from having the user add `TProvider` as a type parameter?

src/TypeShape.TestCases/README.md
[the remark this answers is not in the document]
eiriktsarpalis: I'll see if I can make the changes directly on your branch when I get the time.

eiriktsarpalis:
> You already have an inconsistent package ID format:
> 
> * `typeshape-csharp`
> * `TypeShape.Examples`
> * `TypeShape.Roslyn`
> 
> I went with `TypeShape.TestCases` rather than `typeshape-testcases`.

Yeah.. unfortunately `TypeShape` is taken by the F# library, which is why I'm considering a rename.

PR #50, opened 2024-11-13
Update to .NET 9 sdk

eiriktsarpalis:
cc @AArnott


eiriktsarpalis:
> but targeting .NET 9 as well.

Only for the test, benchmark and demo projects. The NuGet packages remain on 8.

PR #53, opened 2024-11-14
Add nuget.config

Rob-Hague:
On clone+build, I get [NU1507](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1507) because I have multiple package sources defined:

"Warning As Error: There are 3 package sources defined in your configuration. When using central package management, please map your package sources with package source mapping (https://aka.ms/nuget-package-source-mapping) or specify a single package source."

Fix by adding a nuget.config


eiriktsarpalis:
Thank you!

PR #74, opened 2024-12-05
Add support for netstandard2.0

### LINKED ISSUE #49: Support for netstandard2.0
[I have a branch](https://github.com/eiriktsarpalis/PolyType/compare/main...AArnott:PolyType:netstandard2.0) where I'm exploring this now. Besides the "simple" breaks that just require `#if` to avoid undefined APIs, I see these high-level problems:

1. Default interface implementations, such as `IObjectTypeShape` defining
   ```cs
   TypeShapeKind ITypeShape.Kind => TypeShapeKind.Object;
   ```

   This could presumably be resolved by making the default implementation conditional on .NET 8, such that folks supporting .NET Standard 2.0 will have to implement the member themselves. Annoying (for them) but not a big deal. And not a breaking change going from netstandard2.0 to .NET 8 either, I think.
2. static abstract interface members. This is the `IShapeable<T>` elephant in the room.

eiriktsarpalis:
Fix #49.


eiriktsarpalis:
Published as PolyType v0.22

PR #79, opened 2024-12-12
Add support for customizing TypeShape kinds via an attribute.

### LINKED ISSUE #43: Allow setting the TypeShapeKind of a type using an attribute
I'm thinking it might make sense to let users manually pick a shape kind for a type using an attribute annotation when the default convention isn't appropriate.

_Originally posted by @eiriktsarpalis in https://github.com/AArnott/Nerdbank.MessagePack/pull/11#discussion_r1826926632_

eiriktsarpalis:
Fix #43.


eiriktsarpalis:
FYI @AArnott

PR #83, opened 2024-12-13
Add nbgv as a CLI tool

AArnott:
This supports the workflow I suggest in #82

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,0 +1,13 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "nbgv": {
+      "version": "3.7.112",
+      "commands": [
+        "nbgv"
+      ],
+      "rollForward": false
+    }
+  }
+}
--- a/Makefile
+++ b/Makefile
@@ -29,11 +29,10 @@ generate-docs: clean
 	docfx $(DOCS_PATH)/docfx.json
 
 bump-version:
-	test -n "$(VERSION)" || (echo "must specify VERSION" && exit 1)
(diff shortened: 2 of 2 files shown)


eiriktsarpalis:
How does this work? I current use this script to bump versions and generate tags:

https://github.com/eiriktsarpalis/PolyType/blob/fe79cdd7dc2c8150e9e8f576711f398314a4fd3e/Makefile#L31-L36

AArnott:
You'd start with `dotnet tool restore` to get `nbgv` brought down into your repo.

You *could* replace that whole script with:

```
nbgv prepare-release  # prints about how v0.24 stabilization branch is created
nbgv tag v0.24  # to tag that branch for immediate release
```

The first step above will have already incremented your "0.24" in `version.json` to "0.25". The v0.24 branch is what you actually release, and provides a branch for you to make servicing fixes.
The `prepare-release` command is designed for a workflow where your development branch (`main` in many cases) is where most development is. Notice how `main` is immediately bumped for the next version, where you then do development. 
This has two outcomes: every build indicates that it's a prerelease of an upcoming version instead of a successive servicing release of a prior version (that won't ship because you'll bump the version later). It also means that your releases won't tend to end with ".1" but rather ".x".

You certainly don't have to adopt that whole thing. You can keep your existing workflow with this replacement of your script, which should be equivalent, but simpler:

```
test -n "$(VERSION)" || (echo "must specify VERSION" && exit 1) 
nbgv set-version $(VERSION)
git commit -m "Bump version to $(VERSION)"
nbgv tag
```

AArnott:
I updated your `Makefile`. But as I don't have `make` installed, I didn't test it in-context. Though I tested the 3 commands I added out on my CLI.

PR #90, opened 2024-12-18
Add net6.0 to the test targets

eiriktsarpalis:
Even though net6.0 is now out of support, it provides a good way to test the `netstandard2.0` build of the library in non-Windows environments.


tests/PolyType.SourceGenerator.UnitTests/CompilationHelpers.cs
```diff
@@ -235,8 +246,8 @@ public static (int startLine, int startColumn) GetEndPosition(this Location loca
         return (lineSpan.EndLinePosition.Line, lineSpan.EndLinePosition.Character);
     }
 
-#if !NET
-    private const string NetfxPolyfillAttributes = """
+#if !NET || !NET8_0_OR_GREATER
```
eiriktsarpalis, line 249:
```suggestion
#if !NET || NET6_0
```

PR #93, opened 2024-12-25
Use IReadOnlyList when modelling properties and constructor params

### LINKED ISSUE #77: Consider replacing `IObjectTypeShape.GetProperties()` with an `IReadOnlyList<T>`
The existing [`IObjectTypeShape.GetProperties()`](https://eiriktsarpalis.github.io/PolyType/api/PolyType.Abstractions.IObjectTypeShape.html#PolyType_Abstractions_IObjectTypeShape_GetProperties) and [`IConstructorShape.GetParameters()`](https://eiriktsarpalis.github.io/PolyType/api/PolyType.Abstractions.IConstructorShape.html#PolyType_Abstractions_IConstructorShape_GetParameters) methods return an `IEnumerable` of results. This somewhat made sense back when this library only worked on top of reflection and resolving individual properties was expensive, but it creates usability problems for users needing to interrogate whether an object contains properties or not.

The methods should be replaced with properties returning `IReadOnlyList<T>` values instead.

cc @AArnott

eiriktsarpalis:
Fix #77.

### DIFF
--- a/src/PolyType.Examples/CborSerializer/CborSerializer.Builder.cs
+++ b/src/PolyType.Examples/CborSerializer/CborSerializer.Builder.cs
@@ -25,13 +25,11 @@ public static partial class CborSerializer
 
         public object? VisitObject<T>(IObjectTypeShape<T> objectShape, object? state)
         {
-            CborPropertyConverter<T>[] properties = objectShape
-                .GetProperties()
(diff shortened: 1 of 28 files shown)


eiriktsarpalis:
FYI @AArnott this is an upcoming source breaking change.

PR #95, opened 2024-12-27
Have parameter shapes always match the name of their corresponding properties

### LINKED ISSUE #94: Constructor parameter shapes should assume the names of their matching properties
> I still have to (at present) write the code to match a serialized property name to a constructor parameter. This is a pain, and requires that I decide the policy for how user's constructor parameter names must match the property name. Consider this case:
> 
> ```cs
> class Foo
> {
>    public Foo(int a);
> 
>    public int A { get; }
> }
> ```
> 
> By default, this might serialize to JSON like this:
> 
> ```json
> { "A": 5 }
> ```
> 
> Now, upon _de_serialization, how should the deserializer decide what to do with the JSON `A` property? It doesn't match the `a` constructor parameter. So should it drop it, or should it decide based on a camelCase/PascalCase match, or should it be totally case-insensitive?
> 
> From a PolyType perspective, we have an opportunity perhaps to unify how serializers work including potentially some of their policies. Personally, for a fresh MessagePack library such as the one I'm building, I'd rather _not_ have to decide this policy and rely on a common framework like PolyType to define it, so that users can define their types once with PolyType conformance, and then switch from serializing to JSON to MsgPack to Cbor -- and have it all Just Work instead of one of those formats failing because they don't know that property `A` should be matched with ctor parameter `a`.

_Originally posted by @AArnott in https://github.com/eiriktsarpalis/PolyType/issues/44#issuecomment-2466938286_

eiriktsarpalis:
This ensures that constructor parameter shapes assume the names their corresponding properties. A property matches a constructor parameter iff they have the same type and additionally the parameter name is a camel cased variant of the property name.

Fix #94.

FYI @AArnott this should remove the need to manually match parameters to properties.


AArnott:
> the parameter name is a camel cased variant of the property name.

For records with primary constructors, this isn't true. The parameter name will exactly match the property name without a camelcase variance.

eiriktsarpalis:
> > the parameter name is a camel cased variant of the property name.
> 
> For records with primary constructors, this isn't true. The parameter name will exactly match the property name without a camelcase variance.

The identifier comparison is "camel case invariant", see:

https://github.com/eiriktsarpalis/PolyType/blob/22e549426043989b65cd5cd3704cada628120025/src/Shared/Helpers/CommonHelpers.cs#L133-L134

AArnott:
It looks like it will be considered equal regardless of the capitalization of the first letter rather than enforcing a specific camel case conversion, which is good.

PR #96, opened 2025-01-19
Add support for surrogate type shapes.

### LINKED ISSUE #9: Types with a type descriptor are not created properly
Given a type with a type converter, where the type converter is used to correctly create the instance, TSC# does not consider the type converter, instead preferring the default constructor. For example, consider the following type:

```csharp
[TypeConverter(typeof(ApiEndpointUrlTypeConverter))]
public readonly partial struct ApiEndpointUrl
{
    private ApiEndpointUrl(string value)
    {
        Value = value;
    }

    public string Value { get; }
    
    public static ApiEndpointUrl From(string value) => new(value);
}
```

The purpose of which is to stop it being created via `new`. In order to convert it from a string to an instance, for instance, in configuration (like appsettings.json), it has a type converter:

```csharp
internal class ApiEndpointUrlTypeConverter : TypeConverter
{
    public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType) => 
        sourceType == typeof(string) || base.CanConvertFrom(context, sourceType);

    public override object? ConvertFrom(ITypeDescriptorContext? context, CultureInfo? culture, object value) => 
        value is string s ? ApiEndpointUrl.From(s) : base.ConvertFrom(context, culture, value);

    public override bool CanConvertTo(ITypeDescriptorContext? context, Type? sourceType) => 
        sourceType == typeof(string) || base.CanConvertTo(context, sourceType);

    public override object? ConvertTo(ITypeDescriptorContext? context, CultureInfo? culture, object? value, Type destinationType) =>
        value switch
        {
            ApiEndpointUrl aiu when destinationType == typeof(string) => aiu.Value,
            _ => base.ConvertTo(context, culture, value, destinationType)
        };
}
```

The shape generated for this type is something like this:

```csharp
        private global::TypeShape.Abstractions.IConstructorShape CreateConstructor_ApiEndpointUrl()
        {
            return new global::TypeShape.SourceGenModel.SourceGenConstructorShape<global::ApiEndpointUrl, object?>
            {
                DeclaringType = (global::TypeShape.Abstractions.IObjectTypeShape<global::ApiEndpointUrl>)ApiEndpointUrl,
                ParameterCount = 0,
                DefaultConstructorFunc = static () => new global::ApiEndpointUrl(),
                AttributeProviderFunc = static () => typeof(global::ApiEndpointUrl).GetConstructor(InstanceBindingFlags, global::System.Array.Empty<global::System.Type>()),
                IsPublic = true,
            };
        }
```

Maybe there should be a way (a shape?) to signify the  type is decorated with a type converter? 

I'm happy to attempt this if you could validate my idea in the first place.

### LINKED ISSUE #70: Consider adding support for type conversion
Exactly. Technically it does introduce more overhead compared to writing a bespoke converter but at the end of day it isn't substantial and the approach is much more user friendly and generalized.

_Originally posted by @eiriktsarpalis in https://github.com/AArnott/Nerdbank.MessagePack/pull/117#discussion_r1858133651_

@AArnott

eiriktsarpalis:
The `TypeShape` attribute can now be used to specify marshallers to surrogate types:

```C#
[TypeShape(Marshaller = typeof(EnvelopeMarshaller))]
record Envelope(string Value);

class EnvelopeMarshaller : IMarshaller<Envelope, string>
{
    public string? ToSurrogate(Envelope? envelope) => envelope?.Value;
    public Envelope FromSurrogate(string? surrogateString) => new(surrogateString ?? "");
}
```

The above configures `Envelope` to admit a string-based shape using the specified surrogate representation. In other words, it assumes a string schema instead of that of an object. Surrogates are used as a more versatile alternative compared to format-specific converters.

In the following example, we marshal the internal state of an object to a surrogate struct:

```C#
[TypeShape(Marshaller = typeof(Marshaller))]
public class PocoWithInternalState(int value1, string value2)
{
    private readonly _value1 = value1;
    private readonly _value2 = value2;

    public record struct Surrogate(int Value1, string Value2);

    public sealed class Marshaller : IMarshaller<PocoWithInternalState, Surrogate>
    {
        public Surrogate ToSurrogate(PocoWithInternalState? poco) => poco is null ? default : new(poco._value1, poco._value2);
        public PocoWithInternalState FromSurrogate(Surrogate surrogate) => new(poco._value1, poco._value2 ?? "");
    }
}
```

cc @AArnott

Fix #70. Fix #9.

### DIFF
--- a/docs/shape-providers.md
+++ b/docs/shape-providers.md
@@ -78,6 +78,63 @@ ITypeShapeProvider provider = new ReflectionTypeShapeProvider(useReflectionEmit:
 
 PolyType exposes a number of attributes that tweak aspects of the generated shape. These attributes are recognized both by the source generator and the reflection provider.
 
+### TypeShapeAttribute
+
+The `TypeShape` attribute can be applied on type declarations to customize their generated shapes. It is independent of the `GenerateShape` attribute since it doesn't trigger the source generator and is recognized by the reflection-based provider.
+
+The `Kind` property can be used to override the default shape kind for a particular type:
+
+```C#
+[TypeShape(Kind = TypeShapeKind.Object)]
+class MyList : List<int>
+{
+    public int Value { get; set; }
+}
+```
+
+The above will instruct the providers to generate an object shape as opposed to an enumerable shape that is the default. It can also be used to completely disable any nested member resolution for a given type:
+
+```C#
+[TypeShape(Kind = TypeShapeKind.None)]
+record MyPoco(int Value);
+```
+
+#### Surrogate types
+
+The `TypeShape` attribute can also be used to specify marshallers to surrogate types:
+
+```C#
+[TypeShape(Marshaller = typeof(EnvelopeMarshaller))]
+record Envelope(string Value);
+
+class EnvelopeMarshaller : IMarshaller<Envelope, string>
+{
+    public string? ToSurrogate(Envelope? envelope) => envelope?.Value;
+    public Envelope FromSurrogate(string? surrogateString) => new(surrogateString ?? "");
+}
+```
+
+The above configures `Envelope` to admit a string-based shape using the specified surrogate representation. In other words, it assumes a string schema instead of that of an object. Surrogates are used as a more versatile alternative compared to format-specific converters.
+
+In the following example, we marshal the internal state of an object to a surrogate struct:
+
+```C#
+[TypeShape(Marshaller = typeof(Marshaller))]
+public class PocoWithInternalState(int value1, string value2)
+{
+    private readonly _value1 = value1;
+    private readonly _value2 = value2;
+
+    public record struct Surrogate(int Value1, string Value2);
+
+    public sealed class Marshaller : IMarshaller<PocoWithInternalState, Surrogate>
+    {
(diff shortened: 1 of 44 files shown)


AArnott:
Will a shape graph that references `PocoWithInternalState` automatically include the shape of its surrogate type? I don't see any `[GenerateShape]` attributes in your PR description so it's hard to imagine where these might be required.

When my serializer asks for the shape of `PocoWithInternalState` does it get the shape of that type or its surrogate? I'm guessing of the original type, and I need to pay attention to this `TypeShapeAttribute` and do the conversion myself, yes?

AArnott:
Nm. I found the answers to my questions by reading the code changes in your samples.

AArnott:
I'm curious whether there's a story for using surrogate types where the original type is not in user code. For instance, if I wanted to avoid writing a custom converter for `System.IO.Stream` by utilizing a surrogate type, how can I achieve that given I can't add an attribute to `System.IO.Stream`?

eiriktsarpalis:
Great question. It's something I contemplated as well, but decided to postponed it for now (and give the feature itself to prove its worth). Ultimately, it feels like this could be something that could be overridden in witness types via options on the generic `GenerateShape` attribute

PR #97, opened 2025-01-19
Add nullable ref annotations to `IMarshaller<T, TSurrogate>`

AArnott:
In calling into this interface, the lack of these annotations makes callers add additional checks that should not be necessary when we consider the reasonable assumption that a non-null input should produce a non-null output.

### DIFF
--- a/src/PolyType/IMarshaller.cs
+++ b/src/PolyType/IMarshaller.cs
@@ -1,3 +1,5 @@
+using System.Diagnostics.CodeAnalysis;
+
 namespace PolyType;
 
 /// <summary>
@@ -17,6 +19,7 @@ public interface IMarshaller<T, TSurrogate>
     /// </summary>
     /// <param name="value">The input of type <typeparamref name="T"/>.</param>
     /// <returns>A value of type <typeparamref name="TSurrogate"/> corresponding to <paramref name="value"/>.</returns>
+    [return: NotNullIfNotNull(nameof(value))]
     TSurrogate? ToSurrogate(T? value);
 
     /// <summary>
@@ -24,5 +27,6 @@ public interface IMarshaller<T, TSurrogate>
     /// </summary>
     /// <param name="surrogate">The input of type <typeparamref name="TSurrogate"/>.</param>
     /// <returns>A value of type <typeparamref name="T"/> corresponding to <paramref name="surrogate"/>.</returns>
+    [return: NotNullIfNotNull(nameof(surrogate))]
     T? FromSurrogate(TSurrogate? surrogate);
 }


AArnott:
On the other hand, if surrogates are generally expected to be *structs*, even when the original type is a class, what _should_ the marshaler do when it is given a null reference to convert?

eiriktsarpalis:
I'm somewhat on the fence about this. I ended up with the current annotations after quite a bit of experimentation. TL;DR while not perfect for all scenaria, accepting and returning nullable values admits maximal flexibility when it comes to marshalling values. By adding `NotNullIfNotNull` it is guiding marshallers to map non-null values to non-nul values, which might not be appropriate depending on the domain modelling. In a way, it is somewhat equivalent to requiring that inputs and outputs be non-null.

AArnott:
Aside from the annotations then, what do you think of these two points?

> What if the contract was that the marshaler would never be called with a null value, since that requires no surrogate?

and  

> [your own sample](https://eiriktsarpalis.github.io/PolyType/shape-providers.html#surrogate-types) seems to misbehave for null reference types in that passing through the surrogate will always create a non-null instance initialized with default values.

eiriktsarpalis:
> What if the contract was that the marshaler would never be called with a null value, since that requires no surrogate?

You mean the contract of the consuming library or the contract of the marshaller implementation? Given that it's a double-ended abstraction, the library side cannot be certain that null values should be short-circuited for the current type and the marshaller side cannot be certain if null values will be fed into it. My thinking is that the annotation of the marshaller explicitly places the responsibility of null handling on the marshaller's side. FWIW, System.Text.Json tried addressing this problem using the [`JsonConverter.HandleNull`](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonconverter-1.handlenull?view=net-9.0) property that declares its own nullability configuration: in practice it created more problems than it solved because it forced nullable-oblivious annotations on the parameters and confused users on null handling responsibility.

> [your own sample](https://eiriktsarpalis.github.io/PolyType/shape-providers.html#surrogate-types) seems to misbehave for null reference types in that passing through the surrogate will always create a non-null instance initialized with default values.

I wrote the sample somewhat intentionally to demonstrate the flexibility of the abstraction (mapping null to non-null values or vice-versa). If it confuses the message I can try changing it to something a bit more canonical.

AArnott:
> I wrote the sample somewhat intentionally to demonstrate the flexibility of the abstraction (mapping null to non-null values or vice-versa). If it confuses the message I can try changing it to something a bit more canonical.

Well, I found it confusing because it looks unintentional. It seems that mapping a null ref type to a value type, and then back from a value type to a non-null ref type is a "lossy" conversion. I guess that won't be the common intent, so the sample may mislead people to write buggy code. But if you wanted to demonstrate that as a feature, I guess at least call out that that's what's happening, and probably add another sample or just mention how it could be done otherwise.

eiriktsarpalis:
I've submitted https://github.com/eiriktsarpalis/PolyType/pull/99 which updates/fixes the code samples.

PR #98, opened 2025-01-19
Fix error message about private marshalers

AArnott:
Experimentation suggests that marshalers do not in fact need to be `public`. Merely `internal` allows the source generator to access the types.

### DIFF
--- a/src/PolyType.SourceGenerator/Parser/Parser.Diagnostics.cs
+++ b/src/PolyType.SourceGenerator/Parser/Parser.Diagnostics.cs
@@ -81,7 +81,7 @@ public sealed partial class Parser
         title: "Type contains invalid marshaller configuration.",
         messageFormat: 
             "The type '{0}' contains invalid marshaller configuration. " +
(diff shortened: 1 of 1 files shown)


src/PolyType.SourceGenerator/Parser/Parser.Diagnostics.cs
```diff
@@ -81,7 +81,7 @@ public sealed partial class Parser
         title: "Type contains invalid marshaller configuration.",
         messageFormat: 
             "The type '{0}' contains invalid marshaller configuration. " +
-            "A valid marshaller must be a public type with a default constructor and exactly one IMarshaller<,> implementation for the current type.",
+            "A valid marshaller must be an internal or public type with a default constructor and exactly one IMarshaller<,> implementation for the current type.",
```
eiriktsarpalis, line 84: Perhaps we could change it to "accessible" since internal will not work if it's defined in another assembly.
```suggestion
            "A valid marshaller must be an accessible type with a default constructor and exactly one IMarshaller<,> implementation for the current type.",
```

PR #101, opened 2025-01-21
Add `ITypeShapeProvider.GetShape` overload that accepts open generic types

### LINKED ISSUE #100: Support for obtaining closed generic type shapes given an open generic and type arguments
As originally described at https://github.com/AArnott/Nerdbank.MessagePack/issues/182#issuecomment-2565690288, I'd like to enable support for open generic converters while remaining AOT safe in Nerdbank.MessagePack.
The way I propose to do this is by asking the user to use a witness class to close the open generic converter as many times as necessary, and then at runtime the serializer will ask for the closed type shape.

What I need from PolyType is:

1. Add the following member to `ITypeShapeProvider`:
   ```cs
   ITypeShape? GetShape(Type unboundGenericType, ReadOnlySpan<Type> genericTypeArguments);
   ```
2. In the source generated implementation, add a dictionary with a hard-coded content that facilitates implementation of the new method.

AArnott:
Closes #100

### DIFF
--- a/src/PolyType.SourceGenerator/Model/TypeShapeModel.cs
+++ b/src/PolyType.SourceGenerator/Model/TypeShapeModel.cs
@@ -6,6 +6,10 @@ public abstract record TypeShapeModel
 {
     public required TypeId Type { get; init; }
 
+    public TypeId UnboundGenericType { get; init; }
+
+    public ReadOnlyMemory<TypeId> TypeArguments { get; init; }
+
     /// <summary>
     /// A unique identifier deriving from the type name that can be used as a valid member identifier.
     /// </summary>
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,8 @@ indent_style = space
 
 [*.cs]
 
+dotnet_sort_system_directives_first = false
+
 # CA1000: Do not declare static members on generic types
 dotnet_diagnostic.CA1000.severity = none
 
--- a/src/PolyType.SourceGenerator/Parser/Parser.ModelMapper.cs
+++ b/src/PolyType.SourceGenerator/Parser/Parser.ModelMapper.cs
@@ -10,6 +10,14 @@ public sealed partial class Parser
 {
     private TypeShapeModel MapModel(TypeDataModel model, TypeId typeId, string sourceIdentifier)
     {
+        TypeId unboundGenericType = default;
+        ReadOnlyMemory<TypeId> typeArguments = default;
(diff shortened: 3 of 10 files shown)


AArnott:
There are two remaining issues:

1. Tests fail for the case of a nested type of a generic type (e.g. `Outer<T>.Inner`). I'm not sure how to best solve this without going all out and making `TypeShapeModel` refer to containing types with another `TypeShapeModel`. This is what I ultimately did in MessagePack-CSharp in order to represent some types.
2. We may want to add more targeted tests than the test I modified to get some coverage of the newly generated code.

src/PolyType.SourceGenerator/Model/TypeShapeModel.cs
```diff
@@ -6,6 +6,10 @@ public abstract record TypeShapeModel
 {
     public required TypeId Type { get; init; }
 
+    public TypeId UnboundGenericType { get; init; }
+
+    public ReadOnlyMemory<TypeId> TypeArguments { get; init; }
```
eiriktsarpalis, line 11: These types are part of the incremental model, so they need to have structural equality. Because `ROM<T>` uses reference equality, this can result in false negatives when incoming models are being compared. You can use the provided collection types that have the appropriate equality comparison:
```suggestion
    public ImmutableEquatableArray<TypeId> TypeArguments { get; init; }
```

eiriktsarpalis (commented):
So that I can better understand the use case, would it be possible to also implement a generic converter mechanism that uses the new API in one of the examples?

AArnott:
Do you particularly want it in your repo, or would seeing [the sample use case in my own repo's context](https://github.com/AArnott/Nerdbank.MessagePack/commit/fc3fc686f5ea75aed4f534306029e7e45156a33c#diff-bdabc7d935c30b75f7238b041d91eae5633f9fa3ba2ef82589ffcff1b8dc3202) help?

AArnott:
I took a look at your samples, but none of them support the user providing their own custom converter, which is the basis of the problem and my solution. The custom converter may be specified by attribute or at runtime, but the point is that the converter is expressed as an open generic type, which must be closed and activated at runtime since PolyType cannot predict at source generation time what the converter will be, but the developer can, and may be expected to add `[GenerateShape<MyConverter<int>>]` to a witness type.

AArnott:
I wonder if this is an idea worth pursuing. If so, I can perhaps spend more time trying to solve the issues in https://github.com/eiriktsarpalis/PolyType/pull/101#issuecomment-2603644289.

PR #103, opened 2025-01-21
Add ExcludeFromCodeCoverageAttribute to generated ShapeProvider and Witness files

### LINKED ISSUE #102: ExcludeFromCodeCoverage on generated ShapeProvider Types
Can we get  the ``System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage`` attribute added to the generated ShapeProvider Types?
In some of my projects they apparently represent a sufficently large part of the code that some arbitrary PR guard check apparently failed.

Tornhoof:
I've added them only for the ShapeProvider and Witness files, not the partial types annotated with the GenerateShape Attribute as these are under user-control as far as I understand it.

Do you want tests that the attributes are added?
I could not find any tests actually verifying the generated source code or the generated trees.

fixes #102


eiriktsarpalis:
Thanks. I wouldn't be too concerned about testing that.

PR #105, opened 2025-01-23
Extend marshaller support to generic types

eiriktsarpalis:
FYI @AArnott


AArnott:
Awesome. And it appears no change is required in Nerdbank.MessagePack to support this aside from updating the PolyType version.

PR #107, opened 2025-01-26
Add the `AggregatingTypeShapeProvider` class

### LINKED ISSUE #91: Aggregating ITypeShapeProvider
Consider an `ITypeShapeProvider` implementation that aggregates results from a list of other `ITypeShapeProvider` instances.
This aggregator would take a list of other instances and return the first shape from the list of providers that returns a non-null result.
It may include caching of the result for faster repeat calls.

This would enable combining multiple shape providers and querying them in a unified manner.
The scenario I'm thinking about is of an RPC system that may have access to several source generated `ITypeShapeProvider` instances, and may also want to fallback to the reflection-based `ITypeShapeProvider` when no source generated one is known.

## Example usage

```cs
ITypeShapeProvider provider1 = ...;
ITypeShapeProvider provider2 = ...;
ITypeShapeProvider aggregatedProvider = new AggregatedTypeShapeProvider(new[] { provider1, provider2 });

var shape = aggregatedProvider.GetShape(typeof(MyType));
```

AArnott:
There is no cache in the aggregating provider. I could go either way on this, but since I implemented no cache, I also added tests to prove there is no cache.

Closes #91

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <IsPackable>false</IsPackable>
-    <LangVersion>13</LangVersion>
+    <LangVersion>12</LangVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)'))</RepoRoot>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <IsPackable>false</IsPackable>
-    <LangVersion>12</LangVersion>
+    <LangVersion>13</LangVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <RepoRoot>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)'))</RepoRoot>
--- a/src/PolyType/Abstractions/TypeShapeProvider.cs
+++ b/src/PolyType/Abstractions/TypeShapeProvider.cs
@@ -43,4 +43,31 @@ public static class TypeShapeProvider
     /// <exception cref="NotSupportedException"><paramref name="shapeProvider"/> does not support <paramref name="type"/>.</exception>
     public static ITypeShape Resolve(this ITypeShapeProvider shapeProvider, Type type)
         => shapeProvider.GetShape(type) ?? throw new NotSupportedException($"The shape provider '{shapeProvider.GetType()}' does not support type '{type}'.");
+
+    /// <summary>
+    /// Creates a new <see cref="ITypeShapeProvider"/> that returns the first shape found
+    /// from iterating over a given list of providers.
+    /// </summary>
+    /// <param name="providers">The sequence of providers to solicit for type shapes.</param>
+    /// <returns>The aggregating <see cref="ITypeShapeProvider"/>.</returns>
+    public static ITypeShapeProvider Combine(params ReadOnlySpan<ITypeShapeProvider> providers)
+    {
+        return new AggregatingTypeShapeProvider(providers.ToArray());
+    }
+
+    private sealed class AggregatingTypeShapeProvider(ITypeShapeProvider[] providers) : ITypeShapeProvider
+    {
+        public ITypeShape? GetShape(Type type)
+        {
+            foreach (ITypeShapeProvider provider in providers)
+            {
+                if (provider.GetShape(type) is ITypeShape shape)
+                {
+                    return shape;
+                }
+            }
+
+            return null;
+        }
+    }
 }
--- a/src/PolyType/Abstractions/TypeShapeProvider.cs
+++ b/src/PolyType/Abstractions/TypeShapeProvider.cs
@@ -43,31 +43,4 @@ public static class TypeShapeProvider
     /// <exception cref="NotSupportedException"><paramref name="shapeProvider"/> does not support <paramref name="type"/>.</exception>
     public static ITypeShape Resolve(this ITypeShapeProvider shapeProvider, Type type)
         => shapeProvider.GetShape(type) ?? throw new NotSupportedException($"The shape provider '{shapeProvider.GetType()}' does not support type '{type}'.");
-
-    /// <summary>
-    /// Creates a new <see cref="ITypeShapeProvider"/> that returns the first shape found
-    /// from iterating over a given list of providers.
-    /// </summary>
-    /// <param name="providers">The sequence of providers to solicit for type shapes.</param>
-    /// <returns>The aggregating <see cref="ITypeShapeProvider"/>.</returns>
-    public static ITypeShapeProvider Combine(params ITypeShapeProvider[] providers)
-    {
-        return new AggregatingTypeShapeProvider(providers);
-    }
-
-    private sealed class AggregatingTypeShapeProvider(ITypeShapeProvider[] providers) : ITypeShapeProvider
-    {
-        public ITypeShape? GetShape(Type type)
-        {
-            foreach (ITypeShapeProvider provider in providers)
-            {
-                if (provider.GetShape(type) is ITypeShape shape)
-                {
-                    return shape;
-                }
-            }
-
-            return null;
-        }
-    }
 }
--- a/src/PolyType/Abstractions/TypeShapeProvider.cs
+++ b/src/PolyType/Abstractions/TypeShapeProvider.cs
@@ -50,9 +50,9 @@ public static class TypeShapeProvider
     /// </summary>
     /// <param name="providers">The sequence of providers to solicit for type shapes.</param>
     /// <returns>The aggregating <see cref="ITypeShapeProvider"/>.</returns>
-    public static ITypeShapeProvider Combine(params ReadOnlySpan<ITypeShapeProvider> providers)
+    public static ITypeShapeProvider Combine(params ITypeShapeProvider[] providers)
     {
-        return new AggregatingTypeShapeProvider(providers.ToArray());
+        return new AggregatingTypeShapeProvider(providers);
     }
 
     private sealed class AggregatingTypeShapeProvider(ITypeShapeProvider[] providers) : ITypeShapeProvider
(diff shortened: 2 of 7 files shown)


Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <IsPackable>false</IsPackable>
-    <LangVersion>12</LangVersion>
+    <LangVersion>13</LangVersion>
```
eiriktsarpalis, line 4: Is the change using any new language features that make this necessary? I wouldn't object to bumping however C# 12 is the minimum suported baseline that the source generator is targeting, so using that at least for the test types project is a good way to extensive coverage for the source generator.

src/PolyType/Abstractions/TypeShapeProvider.cs
```diff
@@ -43,4 +43,31 @@ public static ITypeShape<T> Resolve<T>(this ITypeShapeProvider shapeProvider)
...
+    /// from iterating over a given list of providers.
+    /// </summary>
+    /// <param name="providers">The sequence of providers to solicit for type shapes.</param>
+    /// <returns>The aggregating <see cref="ITypeShapeProvider"/>.</returns>
+    public static ITypeShapeProvider Combine(params ReadOnlySpan<ITypeShapeProvider> providers)
+    {
+        return new AggregatingTypeShapeProvider(providers.ToArray());
+    }
+
+    private sealed class AggregatingTypeShapeProvider(ITypeShapeProvider[] providers) : ITypeShapeProvider
```
eiriktsarpalis, line 58: One issue we quickly encountered with STJ is that there was a need for users to be able introspect and/or modify the chain of resolvers/providers. A quick way to achieve this is making the aggregating provider public and also have it implement `IReadOnlyList<ITypeShapeProvider>`.

Directory.Build.props
eiriktsarpalis: I presume it's to support `params` span? I think this would render the API unusable in C# 12. Given that it shouldn't be a particularly performance sensitive API I think sticking with param array should be fine.

tests/PolyType.Tests/TypeShapeProviderStaticTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            },
+        };
+        ITypeShapeProvider aggregate = TypeShapeProvider.Combine(first, second);
+        Assert.Same(first.Shapes[typeof(int)], aggregate.GetShape(typeof(int)));
+        Assert.Same(second.Shapes[typeof(string)], aggregate.GetShape(typeof(string)));
+        Assert.Null(aggregate.GetShape(typeof(bool)));
+    }
+
+    /// <summary>
+    /// Validates that the aggregating provider does not cache results.
```
eiriktsarpalis, line 61: While I agree that intermediate providers shouldn't be caching results (caching should be done at level of the consuming visitors), I think it's probably fine to leave as undefined behavior.

Directory.Build.props
AArnott: Yes, it was for the `params` span parameter. And that's fine, I'll switch to an array. 
Strictly speaking though, we might want to (still) copy the array we receive, since the caller can pass in an array and then have mutable access to the same object we use later, leading to undefined behavior. But whether we make that defensive copy should be based on the policy you hold for the library in general. Do you tend to write that off as "you shot your own foot" or do you make the copy? I'll go either way.

src/PolyType/Abstractions/TypeShapeProvider.cs
AArnott: To make it _modifiable_ would require exposing a property too, that is either mutable or references a mutable collection. Mutation opens up thread-safety concerns, which I tend to want to avoid. 

Here's a possible proposal then:
1. Remove the `Combine` method.
2. Expose a public aggregating provider with a public ctor, and a public property for an immutable collection type, but offer a setter so they can replace it in an atomic way.

I would personally prefer that the provider remains immutable, and let the user create a new one and swap out the instance in their own code and deal with the thread-safety ramifications there.
No concerns with allowing them to enumerate the sub-providers though.

tests/PolyType.Tests/TypeShapeProviderStaticTests.cs
AArnott: If we want to document it as undefined, allowing us the freedom to add a cache later, that's... potentially breaking anyway as folks may take dependencies on the old behavior, but I'd go along with it.
But I'd still add tests, so that if the behavior changes, the tests call it out, and we can second guess ourselves and update the tests if necessary, since by that point, it _will_ be an intentional change and having tests that only need a couple changes will help set us up for later too.

src/PolyType/Abstractions/TypeShapeProvider.cs
eiriktsarpalis: I'd prefer keeping it immutable. If modifications become necessary you can always create a fresh aggregated instance.

tests/PolyType.Tests/TypeShapeProviderStaticTests.cs
eiriktsarpalis: Sounds good.

Directory.Build.props
eiriktsarpalis: Defensive copy sounds good.

eiriktsarpalis: FWIW, our approach in the BCL has been to expose both span and array overloads, the idea being that each language version picks the appropriate overload. I wouldn't oppose doing it that way either, but that being said this shouldn't be a hot-path API.

src/PolyType/Utilities/AggregatingTypeShapeProvider.cs
```diff
@@ -0,0 +1,53 @@
+﻿using PolyType.Abstractions;
+
+namespace PolyType.Utilities;
+
+/// <summary>
+/// An <see cref="ITypeShapeProvider"/> that returns the first shape found
+/// from iterating over a given list of providers.
+/// </summary>
```
eiriktsarpalis, line 8: Perhaps name-dropping the chain-of-responsibility pattern in a remarks section would help here.

src/PolyType/Utilities/AggregatingTypeShapeProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        foreach (ITypeShapeProvider provider in providers)
+        {
+            if (provider is null)
+            {
+                throw new ArgumentException("One or more providers are null.", nameof(providers));
+            }
+        }
+
+        _providers = providers;
```
eiriktsarpalis, line 32: Did you skip creating a defensive copy in the end?

Directory.Build.props
AArnott: The lack of a reference to the immutable collections in the netstandard2.0 build of your library keeps surprising me. Is that an intentional choice? If not, I'd tend to use an `ImmutableArray` here, and I could safely expose that directly from the property, allowing for alloc-free enumeration for public users.

src/PolyType/Utilities/AggregatingTypeShapeProvider.cs
AArnott: I was waiting to get your take on it before adding it. I'm adding it in my next iteration.

src/PolyType/Utilities/AggregatingTypeShapeProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+
+        _providers = providers.ToArray();
+    }
+
+    /// <summary>
+    /// Gets the list of providers making up this aggregate provider.
+    /// </summary>
+    public IReadOnlyList<ITypeShapeProvider> Providers => _providers;
```
eiriktsarpalis, line 48: This allows mutation of the underlying array if you cast it. I would recommend changing the type of `_providers` to `IReadOnlyList<ITypeShapeProvider>` and then populate that in the constructor using collection literals:
```C#
_providers = [ ..providers];
```

tests/PolyType.Tests/Utilities/AggregatingTypeShapeProviderTests.cs
```diff
@@ -0,0 +1,179 @@
+﻿using PolyType.Utilities;
+using System.Reflection;
+
+namespace PolyType.Tests.Utilities;
+
+public class AggregatingTypeShapeProviderTests
```
eiriktsarpalis, line 6: Could you also add a test exercising the `IList` property?

src/PolyType/Utilities/AggregatingTypeShapeProvider.cs
eiriktsarpalis: Yeah. It could have used `ImmutableArray<T>` I suppose but maybe it can't just assume that it's available.

AArnott: Ya, I need to reference S.C.I. on net472 to get ImmutableArray. Per [this comment](https://github.com/eiriktsarpalis/PolyType/pull/107#discussion_r1929872467).

However, defining the field to be an interface type means that `foreach` is no longer alloc free. Sure, I could rewrite my `foreach` loop to a `for` loop and use the indexer, but having to remember to avoid `foreach` because the field is typed as an interface instead of a concrete type seems like a fairly precarious position. Generally, and I think there are analyzers to help encourage this, private fields ought to be concrete types to make allocation less likely.

eiriktsarpalis: FWIW I opted for using the compiler-generated IReadOnlyList over ReadOnlyCollection<T> elsewhere in this project. So opting for a dependency on System.Collections.ObjectModel should if anything increase the static footprint of trimmed apps.

tests/PolyType.Tests/Utilities/AggregatingTypeShapeProviderTests.cs
```diff
@@ -0,0 +1,185 @@
...
+    private class MockTypeShapeProvider : ITypeShapeProvider
+    {
+        internal Dictionary<Type, ITypeShape> Shapes { get; } = new();
+
+        public ITypeShape? GetShape(Type type) => Shapes.TryGetValue(type, out ITypeShape? shape) ? shape : null;
+    }
+
+    private class MockTypeShape<T> : ITypeShape<T>
+    {
+        public Type Type => throw new NotImplementedException();
```
eiriktsarpalis, line 167: Not sure why Codecov is giving out coverage reports for test projects -- I thought I'd disabled that.

AArnott: I added an attribute to exclude it.

PR #108, opened 2025-01-26
Base `PrettyPrinter` on `TextWriter` instead of `StringBuilder`

AArnott:
`TextWriter` is a more abstract class, which allows for writing directly to streams or STDOUT, which seems a common scenario for this example, rather than being forced to create one large string.

This doesn't preclude use of `StringBuilder` though, since the `StringWriter : TextWriter` class can create or reuse a `StringBuilder` object while conforming to the `TextWriter` type.


eiriktsarpalis:
Thanks!

PR #110, opened 2025-01-27
Turn off warning-as-error in inner dev loop

### LINKED ISSUE #109: Reduce inner loop friction while holding acceptance criteria to a high standard
While working in your repo, I'm continually slowed by the CS1591 _error_, which requires me to document all public/protected members just to get the code to compile to try out an idea.

I wanted to advocate for a change in policy based on what I found works well in other repos:

1. Use warning-level severity for CS1591 and maybe other diagnostic id's, allowing for quick iterating and testing of an idea.
2. [Turn on warn-as-error in PR/CI builds](https://github.com/AArnott/Library.Template/blob/aa8e87054f2ad8fb64a0de043f39b959a7e5ce51/.github/workflows/build.yml#L49) so that if a pull request comes in that hasn't already resolved all the warnings, it will block the build and thus avoid inadvertently allowing a lower standard of changes to merge.

AArnott:
Closes #109

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ DOCS_PATH := $(SOURCE_DIRECTORY)docs
 CONFIGURATION ?= Release
 NUGET_SOURCE ?= "https://api.nuget.org/v3/index.json"
 NUGET_API_KEY ?= ""
-ADDITIONAL_ARGS ?= -p:ContinuousIntegrationBuild=true
+ADDITIONAL_ARGS ?= -p:ContinuousIntegrationBuild=true -warnAsError -warnNotAsError:NU1901,NU1902,NU1903,NU1904
 CODECOV_ARGS ?= --collect "Code Coverage;Format=cobertura" --results-directory $(ARTIFACT_PATH)
 DOCKER_IMAGE_NAME ?= "polytype-docker-build"
 DOCKER_CMD ?= make CONFIGURATION=$(CONFIGURATION)
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -5,7 +5,6 @@
     <AnalysisLevel>latest-recommended</AnalysisLevel>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsAsErrors>Nullable</WarningsAsErrors>
     <WarningLevel Condition="'$(MSBuildProjectExtension)'!='.fsproj'">9999</WarningLevel>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>


Makefile
```diff
@@ -4,7 +4,7 @@ DOCS_PATH := $(SOURCE_DIRECTORY)docs
 CONFIGURATION ?= Release
 NUGET_SOURCE ?= "https://api.nuget.org/v3/index.json"
 NUGET_API_KEY ?= ""
-ADDITIONAL_ARGS ?= -p:ContinuousIntegrationBuild=true
+ADDITIONAL_ARGS ?= -p:ContinuousIntegrationBuild=true -warnAsError -warnNotAsError:NU1901,NU1902,NU1903,NU1904
```
eiriktsarpalis, line 7: I see the excluded warnings are specific to package vulnerabilities. Is there a particular reason why we'd want to leave those out?

src/Directory.Build.props
```diff
@@ -5,8 +5,6 @@
     <AnalysisLevel>latest-recommended</AnalysisLevel>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <WarningsAsErrors>Nullable</WarningsAsErrors>
```
eiriktsarpalis, line 9: Enforcing nullability warnings as errors seems desirable to me from a modelling perspective. Any chance we could keep this turned on?

eiriktsarpalis (commented):
Given that the motivating use case specifically concerned warnings around XML documentation, could we perhaps selectively disable that family of warnings only during dev time?

Makefile
AArnott: > Is there a particular reason why we'd want to leave those out?

Yes. In my experience, having all PR and CI builds break because of something environmental breaks the flow. When someone "out there" decides a package I'm already using is vulnerable, I want to see that (via dependabot or similar) but I don't want it to stop all work in the meantime.
Also, this repo builds a nuget package. It is not _shipping_ those dependencies so IMO it's not actually that critical. The application that consumes your library will have to update that dependency, and it'll have to do so whether or not your library actually updates it. (if you update it at your repo level, your consumers will have to update your library instead of or in addition to updating the transitive dependency.)

> Is there a way we could enforce the same degree of granularity here?

No, not really. But why would you want to allow warnings in your test projects to creep into your default branch? That just pollutes the VS error list window with warnings until you tune them out. You have a nice clean error list at the moment. You want to keep it that way, don't you?

src/Directory.Build.props
AArnott: I see, you want those as errors even in your inner loop? Sure, I can revert that.

AArnott:
> Given that the motivating use case specifically concerned warnings around XML documentation, could we perhaps selectively disable that family of warnings only during dev time?

CS1591 is just the most common one. In general though, I value the distinction between errors and warnings. Forcing them all to be errors in PR/CI builds is good, but forcing them all to be errors in my inner loop just removes what generally developers see as value to keep distinct.

But it's your repo, and I'd settle for just demoting CS1591 back to a warning in the inner loop if that's as far as you want to go.

AArnott:
@eiriktsarpalis The test failure on Windows appears to be unrelated to this change. Do you want to investigate?

AArnott:
The Windows agent failure vanished on re-run. If you want to investigate for unstable tests, [here was the failed run](https://github.com/eiriktsarpalis/PolyType/actions/runs/12989676387/job/36223176890).

Makefile
eiriktsarpalis: > Also, this repo builds a nuget package. It is not shipping those dependencies so IMO it's not actually that critical.

Agreed. Would it make sense to mute those warnings in general?

eiriktsarpalis:
> The Windows agent failure vanished on re-run. If you want to investigate for unstable tests, [here was the failed run](https://github.com/eiriktsarpalis/PolyType/actions/runs/12989676387/job/36223176890).

It's a nondeterministic failure I've been seeing from time to time in the JSON tests running .NET 6. It appears that the baseline STJ reflection serializer is writing properties in an unexpected order for that runtime only. Given that .NET 6 is now out of support (I only use it to exercise the netstandard2.0 build in non-Windows environments) I'd be inclined to just remove the impacted type for that runtime.

Makefile
AArnott: I'm mixed on that question. My experience with shipping a bunch of OSS packages though tells me that if I don't stay on top of bumping my dependencies for security vulnerabilities, I get (hyper?)-vigilant users filing bugs against the repo. 
Staying current as of each release at least means that when folks consume your package, they're more likely to have a transitive graph that is security-bug-free. So there's value there. IMO it's just not so critical that you should stop all production to bump a package before other PRs can proceed.

AArnott:
Do you want to merge this?

eiriktsarpalis:
I just forgot about it, thanks.

PR #111, opened 2025-02-03
Never emit the `GenerateShapeAttribute<T>` into a user's assembly

AArnott:
It isn't necessary, since it only triggers source generation within the compilation of the code that applies it. It's also problematic because some runtimes (e.g. .NET Framework and unity) do not support generic attributes and reflection APIs throw when they encounter it.


eiriktsarpalis (approved):
Thanks!

PR #123, opened 2025-03-02
Document derived type resolution

### LINKED ISSUE #122: DerivedTypeShapeAttribute docs
The docs you've published leaves me wonder what the behavior is for this case:

```cs
[DerivedTypeShape(typeof(DerivedA))]
partial class Base;

class DerivedA : Base;

class DerivedAB : DerivedA;
```

Now when serializing an `DerivedAB` value, will it be recognized as `Base` (because it is of an unknown type) or as a `DerivedA` (because that is a closer match)?

This behavior must be defined by the delegate returned from `IUnionTypeShape.GetGetUnionCaseIndex()`.
I want to match the behavior in my runtime mapping implementation, so having it documented would be great.

I would say it ideally would be recognized as a `DerivedA`.
In deep trees, this probably means that the type checks in that delegate should be sorted so that the most derived known types are tested first.

eiriktsarpalis:
Fix #122

### DIFF
--- a/docs/shape-providers.md
+++ b/docs/shape-providers.md
@@ -202,6 +202,33 @@ abstract record Node(int label, int left, int right) : BinTree
 If left unset, the name of a derived type defaults to its type name (i.e. `nameof(TDerived)`) and the tag corresponds to the attribute declaration order.
 It should be noted that mono reflection does not preserve attribute declaration ordering, so it is recommended that applications targeting mono should either use the source generator or explicitly set the tags for all model types.
 
+For the case of unregistered derived types, PolyType applies a "nearest ancestor" resolution algorithm. Given the type hierarchy
+
+```C#
+[DerivedTypeShape(typeof(Horse))]
+[DerivedTypeShape(typeof(Cow))]
+class Animal;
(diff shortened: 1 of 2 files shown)


docs/shape-providers.md
```diff
@@ -202,6 +202,33 @@ abstract record Node(int label, int left, int right) : BinTree
...
 It should be noted that mono reflection does not preserve attribute declaration ordering, so it is recommended that applications targeting mono should either use the source generator or explicitly set the tags for all model types.
 
+For the case of unregistered derived types, PolyType applies a "nearest known ancestor" resolution algorithm. Given the type hierarchy
+
+```C#
+[DerivedTypeShape(typeof(Horse))]
+[DerivedTypeShape(typeof(Cow))]
+class Animal;
+class Horse;
+class Cow;
```
AArnott, line 212: Are you missing `: Animal` base types here?

eiriktsarpalis: Thanks, fixed in https://github.com/eiriktsarpalis/PolyType/pull/124

Issue #106, opened 2025-01-24
Extensibility to provide edges from one ITypeSymbol to another for generating type shapes


AArnott:
As a PolyType consumer, Nerdbank.MessagePack has a variety of use cases that require type shapes that PolyType's source generator does not today automatically generate types for. And this goes beyond just needing to apply `[GenerateShape]` on more types.

## Use cases

**Spreading** `IProgress<T>` is an open generic type defined in .NET. StreamJsonRpc provides an open generic converter for this (call it `ProgressConverter<T>`). We need an AOT-safe way to jump from the `ITypeShape<IProgress<T>>` object to an _instance_ of `ProgressConverter<T>`. This does not exist today. #101 is one proposal for how this would work, but it relies on the user manually defining `[GenerateShape<ProgressConverter<int>>]` etc. to ensure that the type shape is available at runtime for every closed type. That's tedious and relies on one bit of code knowing details about the full type graph. It also requires that `ProgressConverter<T>` be a _public_ type (which... maybe we can't escape from anyway, but so far we've kept it `internal`).

## Possible embodiments

- An abstract attribute declared by PolyType that `MessagePackConverterAttribute` derives from, such that PolyType knows that the open generic converter type specified in that attribute should be closed with the same type arguments with which the data type itself is closed. This works for user-defined data types, but cannot apply to `IProgress<T>`.
- An extensibility point in the PolyType source generator that allows us to nominate an assembly for the source generator to load and execute as part of source generation. This assembly will provide a type that implements an interface like this:
  ```cs
  public interface ITypeShapeSuggester
  {
    /// <summary>Suggests additional types for which shapes should be generated, given that a particular type will get a shape for it.</summary>
    IReadOnlyCollection<ITypeSymbol> GetExtraEdges(ITypeSymbol type);
  }
  ```
  An msbuild item type could be fed into the source generator to find these assemblies and load them.
  Nerdbank.MessagePack and StreamJsonRpc could contribute assemblies to this msbuild item list to help PolyType's source generator generate all the types necessary.
  *Retrieving* the type shapes at runtime would still require #101 or similar.

eiriktsarpalis:
> This works for user-defined data types, but cannot apply to IProgress<T>.

Could you clarify why that is?

> I'm willing to build a PoC for the msbuild-based source generator extensibility I proposed above. @eiriktsarpalis are you interested enough that that PoC may be worthwhile?

[...]

PR #128, opened 2025-03-05
Define the public API for the edge from generic shapes to related generic types

AArnott:
#106


src/PolyType/Abstractions/ITypeShape.cs
```diff
@@ -48,4 +48,26 @@ public interface ITypeShape
...
+    /// The generic type definition of the related type (the one found in the <see cref="GenerateShapeEdgeAttribute.To"/> property.
+    /// This open generic type will be closed using the same type arguments used to close the <see cref="Type"/>.
+    /// </param>
+    /// <returns>A factory for the related type, or <see langword="null" /> if no factory for the related type is available.</returns>
+    /// <remarks>
+    /// <see cref="ReflectionProvider.ReflectionTypeShapeProvider"/> can produce the factory on demand without any <see cref="GenerateShapeEdgeAttribute"/>,
+    /// while <see cref="SourceGenModel.SourceGenTypeShapeProvider"/> is expected to only produce the factory that was explicitly requested via attribute.
+    /// </remarks>
+    /// <exception cref="InvalidOperationException">Thrown when this method is called on an instance that does not represent a generic type.</exception>
+    Func<object>? GetRelatedTypeFactory(Type relatedType);
```
AArnott, line 72: Where does this `ITypeShape` object come from? Which assembly defines it? 

Consider this:

- Assembly A defines `ProgressConverter<T>`. It also applies the new assembly attribute linking `IProgress<>` to `ProgressConverter<>`.
- Assembly B defines a data type X with `[GenerateShape]` applied and an `IProgress<int>` property. 
- Assembly C references A and B, and has its own data type Y with `[GenerateShape]` and a property of type X.

At runtime, assembly A eventually gets an ITypeShape for X. It encounters the IPropertyShape whose type is `IProgress<int>` and wants to retrieve the type factory for `ProgressConverter<int>`. 

Will this ITypeShape have the factory? Or will it not because the ITypeShape comes from assembly B, which had no reference to assembly A and thus didn't know the factory would be needed?
Can we somehow have assembly C walk the whole shape graph (even though it reaches into assembly B) and discover `IProgress<int>`, and since C references A, knows that that means an ITypeShape must be generated with the factory for `ProgressConverter<int>`, and somehow make sure that when I'm exploring the shape graph starting from C, A will get it?

src/PolyType/GenerateShapeEdgeAttribute.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// Initializes a new instance of the <see cref="GenerateShapeEdgeAttribute"/> class.
+    /// </summary>
+    /// <param name="from">The generic type definition of a type which, if a type shape is generated for it, should lead to the construction of a factory for an associated generic type.</param>
+    /// <param name="to">
+    /// The generic type definition of the type to make available via <see cref="IGenericTypeShape.GetRelatedTypeFactory(Type)" />.
+    /// This type must be public and have a public, default constructor.
+    /// </param>
+    /// <exception cref="ArgumentNullException"></exception>
+    /// <exception cref="ArgumentException"></exception>
+    public GenerateShapeEdgeAttribute(Type from, Type to)
```
eiriktsarpalis, line 29: I'm guessing this attribute would only have effect for source generators running in the same assembly as the declaration? What would this mean for the reflection provider?

src/PolyType/GenerateShapeEdgeAttribute.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        if (from is null)
+        {
+            throw new ArgumentNullException(nameof(from));
+        }
+
+        if (to is null)
```
eiriktsarpalis, line 36: As a general rule, it's probably best if this type of validation is kept out of attribute types themselves.

src/PolyType/GenerateShapeEdgeAttribute.cs
AArnott: No. The reason it's an assembly-level attribute is to make it efficient for the source generator to search all reference assemblies for this attribute. I need it to impact source generation for all referencing projects as well because it'll be the referencing project that closes `IProgress<>` over some `T`.

src/PolyType/GenerateShapeEdgeAttribute.cs
AArnott: Ya, I figured. Better to put it in analyzers so the errors will be caught at compile-time. 
But this was an easy way to document my expectations of allowed values so you can observe them in the draft PR.

src/PolyType/GenerateShapeEdgeAttribute.cs
```diff
@@ -0,0 +1,63 @@
...
+            throw new ArgumentException(GenericTypeDefinitionRequired, nameof(to));
+        }
+
+        if (from.GetGenericArguments().Length != to.GetGenericArguments().Length)
+        {
+            throw new ArgumentException("The number of generic arguments must match between the two types.");
+        }
+
+        From = from;
+        To = to;
```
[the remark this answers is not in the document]
AArnott: Yes, as documented via xml doc comment on the parameter:

> This type must be public and have a public, default constructor.

src/PolyType/Abstractions/ITypeShape.cs
AArnott: I quick test shows that in fact type shapes are _not_ shared across assemblies. When assembly A defines a shape for type A1, and assembly B defines a shape for type B2, which references A1, the source generator for assembly B emits code for type shapes for both A1 and B2. It does _not_ reuse A1 from assembly A.

So while that seems less efficient than ideal, it _does_ guarantee that the shape provider for B will be the 'provider' for `ITypeShape<A1>`, and thus if the provider is consulted for related types using this API, it should work! :)

AArnott:
The latest push produces type shapes that, if they have related types via attributes, look like this:

```
    15:         private global::PolyType.Abstractions.ITypeShape<global::GenericClass<int, string>> __Create_GenericClass_Int32_String()
    16:         {
    17:             return new global::PolyType.SourceGenModel.SourceGenObjectTypeShape<global::GenericClass<int, string>>
    18:             {
    19:                 CreateConstructorFunc = __CreateConstructor_GenericClass_Int32_String,
    20:                 IsRecordType = false,
    21:                 IsTupleType = false,
    22:                 Provider = this,
    23:                 RelatedTypeFactories = static relatedType => relatedType == typeof(global::GenericConverter<,>) ? () => new global::GenericConverter<int, string>() : null,
    24:             };
    25:         }
```

The mechanism by which I do it, and the form it takes here, can be improved, of course. But it demonstrates the concept. This allows me to jump from an `ITypeShape` to a new instance of a related type.
This particular syntax will likely load all the related types the first time it is invoked. Getting the Type.FullName and switching on the string instead of the type would avoid that, but the cost of a string comparison is probably higher than the cost of comparing a Type. Which one is better? I don't know. But the public API will allow either to be implemented, so the optimization is an implementation detail.

src/PolyType/GenerateShapeEdgeAttribute.cs
eiriktsarpalis: Could this potentially be a property on the `GenerateShape<T>` attribute instead? It's the de facto standard entrypoint for configuring types you don't own, so I would expect any further out-of-band customizability to be going there. Effectively, `GenerateShape<T>` should be reflecting most or all properties present on `TypeShapeAttribute`.

AArnott: > Could this potentially be a property on the `GenerateShape<T>` attribute instead? 

I don't think so. Where would this attribute appear? Not in the `IProgress<T>` assembly (because that's mscorlib)`. Not in the `ProgressConverter<T>` assembly, because there is no shape being generated there. Not in the assembly that declares the shape that includes `IProgress<int>` either, since they don't know about `ProgressConverter<T>`, and _may_ not even be aware that `IProgress<int>` appears in their object graph. When they introduce `IProgress<bool>` into their serialized object graph, it shouldn't be broken by default until they see the runtime exception and then go and add an attribute on a witness type. If I were willing to make them do that, we wouldn't need anything from PolyType at all because Nerdbank.messagepack already exposes a `RegisterConverter` method by which they could add a runtime `ProgressConverter<bool>` type. That in fact is what StreamJsonRpc does in its topic branch, but we consider it to be an unacceptable usability loss from all the other serialization libraries.

eiriktsarpalis: I see. So the use case basically is injecting configuration for generic types that _could_ appear in some third party type graph. My gut feeling tells me that this feature should require that the `to` type must be in the same assembly as the attribute declaration, otherwise we open ourselves up to ambiguity introduced by orphaned declarations in unrelated assemblies that might be present in the application (and therefore resolvable by the reflection provider) but missing from the compile-time source generation scope.

Am I correct in thinking that this scenario is closely related to the "reverse surrogates one"?

AArnott: > this feature should require that the `to` type must be in the same assembly as the attribute declaration

I have no objection to this constraint, as all the use cases I have in mind will fit neatly within it. And if we learn of a use case outside that, we can always revisit then.

> Am I correct in thinking that this scenario is closely related to the "reverse surrogates one"?

I think so, at least in that both attach data to an existing type's shape, possibly from another assembly. And also in the respect that they may only impact shape graphs that were built while referencing an assembly that defines the surrogate.

src/PolyType.Examples/ObjectMapper/Mapper.Builder.cs
```diff
@@ -376,6 +376,7 @@ private sealed class MapperShape<TSource, TTarget>(ITypeShape<TSource> source, I
         public Type Type => typeof(Mapper<TSource, TTarget>);
         public ICustomAttributeProvider? AttributeProvider => typeof(Mapper<TSource, TTarget>);
         public object? Accept(ITypeShapeVisitor visitor, object? state = null) => ((MapperTypeShapeVisitor)visitor).VisitMapper(source, target, state);
+        public Func<object>? GetRelatedTypeFactory(Type relatedType) => throw new NotImplementedException();
```
eiriktsarpalis, line 379: I'm not sure how I feel about use of the term "related" here. A few other alternatives:

1. "associated"
2. "auxiliary"
3. "sidecar"

eiriktsarpalis: Should this be exposed as `IReadOnlyDictionary<Type, Func<object>> AssociatedTypes`? for easier enumeration?

AArnott: I don't think we want folks enumerating what the associated types are. Besides it being none of their business (if they can't predict what the associated type is, they shouldn't be using it), that would also require that we _load_ all those types and the assemblies that declare them.

eiriktsarpalis: I'm still on the fence about keying the factories on the types of the generic definition. If I'm understanding the behavior correctly, you look up a `typeof(ProgressConverter<>)` and get back a factory for `ProgressConverter<Foo>` but if you look up `typeof(ProgressConverter<Foo>)` you get back nothing. I get the motivation but it does feel like a POLA violation.

From the perspective of consuming this method, needing to know that the converter type is specifically `ProgressConverter<>` seems like an implementation detail. Couldn't this be made to work with `typeof(MessagePackConverter<IProgress<Foo>>)` keys instead? From a codegen perspective the method could look as follows:

```C#
static (Type key) =>
{
    if (key.IsAssignableFrom(typeof(GenericConverter<Foo>))
    {
         return () => new GenericConverter<Foo>();
    }

    if (key.IsAssignableFrom(typeof(SomeOtherGenericConverter<Foo>))
    {
        return () => new SomeOtherGenericConverter<Foo>();
    }

    return null;
}
```

This should result in simpler consumer-side semantics while also accommodating ambiguity -- if more than one converter is specified per type this can be accounted for specializing the key type.

eiriktsarpalis: Secondarily, I wonder if returning returning a factory is strictly speaking necessary. Given that the feature is geared towards converters, the lifetime of these objects are typically singletons and therefore returning a fast factory is not a strong prerequisite.

src/PolyType/TypeShapeExtensionAttribute.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// <remarks>
+    /// <para>
+    /// Each type must declare a public default constructor.
+    /// </para>
+    /// <para>
+    /// If <see cref="Target"/> is a generic type definition, each type in this array must also be a generic type definition
+    /// with the same number of generic type parameters.
+    /// </para>
+    /// </remarks>
+    public Type[] AssociatedTypes { get; init; } = [];
```
eiriktsarpalis, line 45: In the past, we discussed a similar feature that lets users include associated _shapes_ in the generated type graph. To ensure we future-proof this API as much as possible, how would we name that property in this attribute? Is it possible that the two could be integrated? (I suspect that the answer is no).

src/PolyType.Examples/ObjectMapper/Mapper.Builder.cs
AArnott: > look up a `typeof(ProgressConverter<>)` and get back a factory for `ProgressConverter<Foo>` but if you look up `typeof(ProgressConverter<Foo>)` you get back nothing.

I'm flexible on this point. Especially given we now support non-generic associated types, I'm happy to make the change to allow specifying a closed generic type and getting back a factory (provided of course that the type args match).

> From the perspective of consuming this method, needing to know that the converter type is specifically `ProgressConverter<>` seems like an implementation detail. Couldn't this be made to work with `typeof(MessagePackConverter<IProgress<Foo>>)` keys instead?

Remember this is actually crucial to uniqueness. It is perfectly legit to have multiple `IProgress<>` converters as associated types. If you ask for `MessagePackConverter<IProgress<>>`, an ambiguity would arise.
It is the serializer's business to know which type they actually want (e.g. an `MyParticularConverter<>`) and by asking for it, there is never any ambiguity. At least for my scenario, it's no simpler yet more problematic to use the base type of an associated type.
Remember as well that I expect we'll perf-optimize this to render the Type we get in this method to a string (using `.AssemblyQualifiedFullName`, presumably) so that unwanted types and their assemblies are not loaded. That precludes assignability checks.

> I wonder if returning returning a factory is strictly speaking necessary

For my use case, once I get the factory I immediately execute it, and discard the factory. So a factory isn't necessary for my use case. If you'd prefer to just return a new instance of the object each time it's invoked, that's fine. If some new use case requires new objects over time, they can always return to the method that creates them.

src/PolyType/TypeShapeExtensionAttribute.cs
AArnott: Yes, I'd initially proposed that we creates shapes for these things, because shapes include constructor delegates. But you pointed out that shapes include far more than we need, and may even follow a trail of members to include many more shapes, which would bloat the application. So I don't think we'd ever want to _combine_ these things.

We _could_ have an `AssociatedShapes` property next to this one for folks who want a full-on shape.

src/PolyType.Examples/ObjectMapper/Mapper.Builder.cs
eiriktsarpalis: > Remember this is actually crucial to uniqueness. It is perfectly legit to have multiple `IProgress<>` converters as associated types. If you ask for `MessagePackConverter<IProgress<>>`, an ambiguity would arise.

How does a library allowing multiple `IProgress<>` pick one in the case of ambiguity. Assuming it first looks for `ProgressConverter<>` and then falls back to something else, it seems that the above codegen scheme should cover that. You use a more derived type as key first, then fall back to the more general one.

AArnott: > How does a library allowing multiple `IProgress<>` converters pick one in the case of ambiguity? 

In my world, it isn't for a library to 'allow' multiple. It only allows one. But users specify which one. All others are out of the serializer's world of concern and the serializer doesn't keep a list of possible candidates.

You may recall from yesterday's design discussion this syntax:

```cs
serializer.AddConverter(new SomeConverter());
```

This is very typical to add custom converters across many serializers, though the exact method name may vary.
In fact I cannot think of a single serializer that can automatically find a custom converter defined by the user *unless* the original data type points to it (often by an attribute). All this already works in Nerdbank.MessagePack. And so do _closed_ generic converters:

```cs
serializer.AddConverter(new SomeConverter<int>());
```

MesagePack-CSharp also supports _open_ generic converters, but it does so in a non-AOT safe way. And it requires users to write a whole class (called a resolver) that is responsible for creating the converter, typically via reflection.

What I'm going for here is an AOT safe and easier variant. It allows attributes or runtime registration of converters, just like all or most other serializers do.

```
[MessagePackConverter(typeof(SomeConverter<>))]`
// or
serializer.AddConverter(typeof(SomeConverter<>));
```

So just like all other serializers, the particular type (not some base type) is brought directly to the attention of the serializer. All I need to do is activate it. There is no ambiguity. But in the case of open generics, AOT safety requires an assist, which is what this PR is all about.

To allow activation by a base class is not at all necessary or desirable from the perspective of a serializer. But it _is_ very much desirable to avoid loading assemblies unnecessarily.

AArnott: I'll prepare the change now with the perf improvement that I'm certain will be insisted on when Visual Studio picks this up, which may help illustrate why it's incompatible with the base type request improvement you're suggesting.

AArnott: I'll also fix it to accept closed generic type requests at the same time, since you suggested it and it isn't incompatible with other objectives.

AArnott: I've pushed an update that uses strings to avoid unwanted assembly loads, and support for acquiring by _closed_ generic types. Acquiring by generic type definition of course still works, since that's crucial to my scenario.

docs/shape-providers.md
```diff
@@ -164,6 +164,30 @@ public record MyPoco<T>(T Value)
...
+#### Associated types
+
+The `TypeShape` attribute can also be used to specify associated types for the target type.
+This provides a Reflection-free way to activate an associated type via its default constructor
+given its @PolyType.Abstractions.ITypeShape.
+For example, serializers may need to jump from a type shape to its converter.
+
+[!code-csharp[](../samples/CSharpSamples/AssociatedTypes.cs#TypeShapeOneType)]
+
+@PolyType.Abstractions.ITypeShape.GetAssociatedTypeFactory*?displayProperty=nameWithType is the method to use to obtain a factory for instances of the associated type.
```
eiriktsarpalis, line 176: Is this syntax for linking to API docs generated by docfx? I didn't that was possible, pretty cool.

docs/shape-providers.md
```diff
@@ -164,6 +164,30 @@ public record MyPoco<T>(T Value)
...
+
+An associated type must have a public default constructor.
+An associated type must have at least `internal` visibility to activate the associated type from within its same assembly,
+but making the type `public` is highly recommended so that when the data type is used in other assemblies, the associated type can be activated from their context as well.
+
+Registering associated types is particularly important when the associated type is generic, and the generic type arguments come from the target type.
```
eiriktsarpalis, line 182: My understanding is that generic type specialisation is the only use case. This feature shouldn't be necessary for non-generic types because you can use the `DynamicDependency` attribute to ensure that the type is always constructible. As such, I would promote this sentence to the start of section.

eiriktsarpalis: Thinking out loud a bit here, given that `DynamicallyAccessedMembers.PublicParameterlessConstructor` provides a built-in mechanism for ensuring constructibility even for generic types, I wonder if all `GetAssociatedType` method needs to return is a `Type` instance representing the relevant specialization: you give it `typeof(ProgressConverter<>)` and it gives you back `typeof(ProgressConverter<Foo>)` which should just work with `Activator.CreateInstance` provided it has been annotated appropriately.

This should also generalize the mechanism beyond just factories and just offer a safe replacement `Type.MakeGenericType`.

samples/CSharpSamples/AssociatedTypes.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace CSharpSamples;
+
+public partial class AssociatedTypes
```
eiriktsarpalis, line 3: If this project is meant to feed into testing, it might make sense to move under the `tests/` folder -- `samples/` is for showcasing demos.

src/PolyType.Roslyn/Helpers/RoslynHelpers.cs
```diff
@@ -220,16 +220,16 @@ type is
     public static bool HasSetsRequiredMembersAttribute(this IMethodSymbol constructor)
     {
         return constructor.MethodKind is MethodKind.Constructor &&
-            constructor.GetAttributes().Any(attr =>
```
eiriktsarpalis, line 223: Ugh. I need to turn on the relevant analyzer for trailing whitespace but keep forgettig to do that.

src/PolyType.Roslyn/KnownSymbols.cs
```diff
@@ -19,6 +24,18 @@ public class KnownSymbols(Compilation compilation)
...
+    /// <summary>
+    /// Gets the symbol for TypeShapeExtensionAttribute.
+    /// </summary>
+    public INamedTypeSymbol TypeShapeExtensionAttribute => _TypeShapeExtensionAttribute ??= Compilation.GetTypeByMetadataName("PolyType.TypeShapeExtensionAttribute") ?? throw new InvalidOperationException("TypeShapeExtensionAttribute not found.");
+    private INamedTypeSymbol? _TypeShapeExtensionAttribute;
+
+    /// <summary>
+    /// Gets the symbol for AssociatedTypeAttributeAttribute.
+    /// </summary>
+    public INamedTypeSymbol AssociatedTypeAttributeAttribute => _AssociatedTypeAttributeAttribute ??= Compilation.GetTypeByMetadataName("PolyType.AssociatedTypeAttributeAttribute") ?? throw new InvalidOperationException("AssociatedTypeAttributeAttribute not found.");
```
eiriktsarpalis, line 36: Why not just make nullable like with the other known symbols in this class?

src/PolyType.Roslyn/KnownSymbols.cs
```diff
@@ -19,6 +24,18 @@ public class KnownSymbols(Compilation compilation)
     public IAssemblySymbol CoreLibAssembly => _CoreLibAssembly ??= Compilation.GetSpecialType(SpecialType.System_Int32).ContainingAssembly;
     private IAssemblySymbol? _CoreLibAssembly;
 
+    /// <summary>
+    /// Gets the symbol for TypeShapeExtensionAttribute.
+    /// </summary>
+    public INamedTypeSymbol TypeShapeExtensionAttribute => _TypeShapeExtensionAttribute ??= Compilation.GetTypeByMetadataName("PolyType.TypeShapeExtensionAttribute") ?? throw new InvalidOperationException("TypeShapeExtensionAttribute not found.");
```
eiriktsarpalis, line 30: I generally try to keep all PolyType runtime attributes in the `PolyType.SourceGenerator` project and not in `PolyType.Roslyn` since the latter is supposed to be a library-agnostic type graph traversal component.

src/PolyType.Roslyn/Model/TypeExtensionModel.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.CodeAnalysis;
+using System.Collections.Immutable;
+
+namespace PolyType.Roslyn;
+
+/// <summary>
+/// Captures the data collected from all the TypeShapeExtensionAttribute attributes that target a particular type.
+/// </summary>
+public sealed record TypeExtensionModel
```
eiriktsarpalis, line 9: Likewise, this would be more of a `TypeShape.SourceGenerator` concept as opposed to a `TypeShape.Roslyn` concept. What I've been doing for things like surrogates is define extensions on the `TypeShape.Roslyn` model types in the source generator alongside the incremental model:

https://github.com/eiriktsarpalis/PolyType/blob/24ea2d1adc44c79af2a0d3e17f0d24ac16e6854e/src/PolyType.SourceGenerator/Model/SurrogateShapeModel.cs#L12-L16

src/PolyType.SourceGenerator/Parser/Parser.Diagnostics.cs
```diff
@@ -110,4 +110,36 @@ public sealed partial class Parser
         category: "PolyType.SourceGenerator",
         defaultSeverity: DiagnosticSeverity.Error,
         isEnabledByDefault: true);
+
+    private static DiagnosticDescriptor AssociatedTypeInternal { get; } = new DiagnosticDescriptor(
+        id: "TS0014",
+        title: "Associated type is internal.",
+        messageFormat: "The associated type '{0}' is internal, which will prevent its activation from within a [GenerateShape] graph produced from another assembly. Associated types should generally be public.",
```
eiriktsarpalis, line 117: That should hold true for pretty much anything within a `[GenerateShape]` type graph that is marked internal, but we generally don't warn about it. Perhaps it makes sense to only emit a diagnostic in cases where the source generator cannot access the associated type?

src/PolyType.SourceGenerator/Parser/Parser.Diagnostics.cs
```diff
@@ -110,4 +110,36 @@ public sealed partial class Parser
...
+        title: "Associated type arity mismatch.",
+        messageFormat: "The associated type '{0}' has arity {1} which must be 0 or match the target type, which has arity {2}.",
+        category: "PolyType.SourceGenerator",
+        defaultSeverity: DiagnosticSeverity.Error,
+        isEnabledByDefault: true);
+
+    private static DiagnosticDescriptor AssociatedTypeInExternalAssembly { get; } = new DiagnosticDescriptor(
+        id: "TS0017",
+        title: "Associated type must be declared in same assembly.",
+        messageFormat: "The associated type '{0}' is not declared in the same assembly as required.",
```
eiriktsarpalis, line 141: I presume you mean this?
```suggestion
        messageFormat: "The associated type '{0}' is not declared in the same assembly as the attribute annotation declaring it.",
```

src/PolyType.SourceGenerator/Parser/Parser.ModelMapper.cs
```diff
@@ -16,13 +18,93 @@ private TypeShapeModel MapModel(TypeDataModel model, TypeId typeId, string sourc
...
+                {
+                    ReportDiagnostic(AssociatedTypeInaccessibleError, location, openType.GetFullyQualifiedName());
+                    return false;
+                }
+
+                return true;
+            }
+        }
+
+        ImmutableEquatableArray<AssociatedTypeId> associatedTypes = associatedTypesBuilder.ToImmutableEquatableArray();
```
eiriktsarpalis, line 98: Feels like this could all be extracted into a helper function returning the associated types.

src/PolyType.SourceGenerator/Parser/Parser.cs
```diff
@@ -20,31 +20,96 @@ public sealed partial class Parser : TypeDataModelGenerator
         CommonHelpers.CreateTupleComparer<ITypeSymbol, string>(
             SymbolEqualityComparer.Default,
             CommonHelpers.CamelCaseInvariantComparer.Instance);
-    
+
     private readonly PolyTypeKnownSymbols _knownSymbols;
+    private readonly ImmutableDictionary<INamedTypeSymbol, TypeExtensionModel> _typeShapeExtensions;
```
eiriktsarpalis, line 25: Do we need to use ImmutableDictionary? It's not the fastest collection and I presume we're not doing persistable updates either given that the field is readonly.

src/PolyType.SourceGenerator/Parser/Parser.cs
```diff
@@ -20,31 +20,96 @@ public sealed partial class Parser : TypeDataModelGenerator
...
         }
 
-        Parser parser = new(knownSymbols.Compilation.Assembly, knownSymbols, cancellationToken);
+        Parser parser = new(knownSymbols.Compilation.Assembly, typeShapeExtensions, knownSymbols, cancellationToken);
         TypeDeclarationModel shapeProviderDeclaration = CreateShapeProviderDeclaration(knownSymbols.Compilation);
         ImmutableEquatableArray<TypeDeclarationModel> generateShapeTypes = parser.IncludeTypesUsingGenerateShapeAttributes(generateShapeDeclarations);
         return parser.ExportTypeShapeProviderModel(shapeProviderDeclaration, generateShapeTypes);
     }
 
+    public static ImmutableArray<TypeExtensionModel> DiscoverTypeShapeExtensions(KnownSymbols knownSymbols, Compilation compilation, MetadataReference metadataReference, CancellationToken cancellationToken)
```
eiriktsarpalis, line 51: Should this be private?

src/PolyType.SourceGenerator/PolyTypeGenerator.cs
```diff
@@ -21,13 +22,30 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
         IncrementalValueProvider<PolyTypeKnownSymbols> knownSymbols = context.CompilationProvider
             .Select((compilation, _) => new PolyTypeKnownSymbols(compilation));
 
+        IncrementalValuesProvider<TypeExtensionModel> externalTypeShapeExtensions = context.MetadataReferencesProvider
+            .Combine(context.CompilationProvider)
+            .Combine(knownSymbols)
+            .SelectMany((tuple, token) => Parser.DiscoverTypeShapeExtensions(tuple.Right, tuple.Left.Right, tuple.Left.Left, token));
+
+        IncrementalValuesProvider<TypeExtensionModel> ownTypeShapeExtensions = context.CompilationProvider
```
eiriktsarpalis, line 30: Can these two incremental values be fused into just one? It seems they both depend on MetadataReferencesProvider and CompilationProvider.

eiriktsarpalis: I'm guessing it has something to do with the logic starting in line 34. Assuming there is a way we could fuse some of this logic so that the source generator defines a smaller incremental graph overall that would be great.

eiriktsarpalis: The more I look at this the more I think it can be simplified. The `Compilation` objects already encapsulates all metadata references via the [Compilation.References property](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.compilation.references?view=roslyn-dotnet-4.9.0) so to me it seems all this logic can be removed from the incremental graph and hidden inside the `Parser` logic.

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.Object.cs
```diff
@@ -39,6 +41,25 @@ private void FormatObjectTypeShapeFactory(SourceWriter writer, string methodName
...
+        if (objectShapeModel.AssociatedTypes.Length == 0)
+        {
+            return "null";
+        }
+
+        StringBuilder builder = new();
+        builder.Append("static associatedType => associatedType switch { ");
+        foreach (AssociatedTypeId associatedType in objectShapeModel.AssociatedTypes)
+        {
+            builder.Append($"\"{associatedType.Open}\" or \"{associatedType.Closed}\" => () => new {associatedType.Constructor}(), ");
```
eiriktsarpalis, line 55: Nit: use triple-quote strings to avoid string escaping.

src/PolyType.SourceGenerator/Parser/Parser.ModelMapper.cs
```diff
@@ -2,6 +2,8 @@
 using PolyType.Roslyn;
 using PolyType.SourceGenerator.Helpers;
 using PolyType.SourceGenerator.Model;
+using System.Collections.Concurrent;
```
eiriktsarpalis, line 5: Is Concurrent necessary?

src/PolyType/ReflectionProvider/ReflectionTypeShape.cs
```diff
@@ -13,4 +16,19 @@ internal abstract class ReflectionTypeShape<T>(ReflectionTypeShapeProvider provi
     ITypeShapeProvider ITypeShape.Provider => provider;
     ICustomAttributeProvider? ITypeShape.AttributeProvider => typeof(T);
     object? ITypeShape.Invoke(ITypeShapeFunc func, object? state) => func.Invoke(this, state);
+
+    public Func<object>? GetAssociatedTypeFactory(Type relatedType)
+    {
+        if (relatedType.IsGenericTypeDefinition && this.Type.GenericTypeArguments.Length != relatedType.GetTypeInfo().GenericTypeParameters.Length)
+        {
+            throw new ArgumentException($"Related type arity ({relatedType.GenericTypeArguments.Length}) mismatch with original type ({this.Type.GenericTypeArguments.Length}).");
```
eiriktsarpalis, line 24: I take it that the attribute annotations have no effect in the case of reflection? Can we document this?

docs/shape-providers.md
AArnott: Yes, that is one of [several allowed forms](https://dotnet.github.io/docfx/docs/links-and-cross-references.html#markdown-links).

samples/CSharpSamples/AssociatedTypes.cs
AArnott: This isn't meant for executable tests, no. It's used directly by your docfx site. Instead of typing code samples in markdown files, you can type it into compiled .cs files to prove that they are and remain valid (and refactorings fix it automatically). That's why there are multiple `#region`s in this file -- the markdown references which one to inline into a particular location.

src/PolyType.Roslyn/KnownSymbols.cs
AArnott: IIRC Copilot suggested this to me, and I assumed it was because this was the style used in your file. I'll fix to conform.

src/PolyType.Roslyn/KnownSymbols.cs
AArnott: I'll move it.

src/PolyType.SourceGenerator/Parser/Parser.ModelMapper.cs
```diff
@@ -16,13 +18,93 @@ private TypeShapeModel MapModel(TypeDataModel model, TypeId typeId, string sourc
...
+        ITypeSymbol[] typeArgs = namedType?.GetRecursiveTypeArguments() ?? [];
+        foreach ((INamedTypeSymbol openType, IAssemblySymbol associatedAssembly, Location? location) in associatedTypeSymbols)
+        {
+            if (!SymbolEqualityComparer.Default.Equals(openType.ContainingAssembly, associatedAssembly))
+            {
+                ReportDiagnostic(AssociatedTypeInExternalAssembly, location, openType.GetFullyQualifiedName());
+                continue;
+            }
+
+            if (!this.IsAccessibleSymbol(openType))
```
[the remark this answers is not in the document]
AArnott: I'll fix it. Another style analyzer you can activate to help enforce. :) Maybe I'll follow up with a PR to turn on some more style rules.

src/PolyType.SourceGenerator/Parser/Parser.cs
AArnott: I'll change it to `IReadOnlyDictionary`

src/PolyType.SourceGenerator/Parser/Parser.cs
AArnott: No. It's used in PolyTypeGenerator. It could be `internal`, but I don't see you using that very much in these projects, so I was trying to match your style.

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.Object.cs
AArnott: I could. But then I'd need to put the content on its own line (especially because the string itself starts with a quote character. And then your the trim trailing whitespace rule would remove the last space on the line, slightly degrading the generated C# code.

src/PolyType.SourceGenerator/Parser/Parser.ModelMapper.cs
AArnott: Nope. Removed.

src/PolyType/ReflectionProvider/ReflectionTypeShape.cs
AArnott: Sure. I expect you mean in shape-providers.md, where you've asked for some revisions already. I'll mix this one in.

src/PolyType.SourceGenerator/PolyTypeGenerator.cs
AArnott: Sounds good. I added this before I understood that the Parser was doing everything on every compilation anyway.

src/PolyType.SourceGenerator/Parser/Parser.Diagnostics.cs
AArnott: Ok, yes I see `TypeNotAccessible` is there. I'll delete this and use that.

docs/shape-providers.md
AArnott: > My understanding is that generic type specialisation is the only use case. This feature shouldn't be necessary for non-generic types because you can use the DynamicDependency attribute to ensure that the type is always constructible. As such, I would promote this sentence to the start of section.

True. IIRC you were the one to call out that it ought to work for non-generics as well. But promoting its use for generics makes total sense.

>  I wonder if all GetAssociatedType method needs to return is a Type instance representing the relevant specialization: you give it `typeof(ProgressConverter<>)` and it gives you back `typeof(ProgressConverter<Foo>)` which should just work with Activator.CreateInstance provided it has been annotated appropriately.

I'm satisfied that this would work. Here's a sample of the updated generated method:

```cs
[return: DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
Type? CloseThis(string openGenericName)
{
	if (openGenericName == "ListOfTimeOnly")
	{
		return ReturnTypeOnly();
	}

	return null;

	[MethodImpl(MethodImplOptions.NoInlining)]
	[return: DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
	static Type ReturnTypeOnly() => typeof(List<TimeOnly>);
}
```

Note the use of method attributes and local functions to retain avoidance in loading assemblies for unwanted types.
This will still accommodate the use cases I have in mind. So I'll prepare this change.

AArnott:
All suggestions resolved except for the switch from an object factory to a function that simply closes generic types. I'll probably do that tomorrow or Monday.

docs/shape-providers.md
eiriktsarpalis: `DynamicallyAccessedMemberTypes.PublicParameterlessConstructor` is specific to the converters scenario, but other scenarios might require other activation or access patterns. Wouldn't it suffice for the attribute to be present on the generic type definition itself?

My feeling is that there is an inherent tension between this feature being just a proxy for `Type.MakeGenericType` and extracting a full-blown typeshape for the associated type. Making the feature only keep the default constructor but discarding everything else is motivated by the use case at hand, but there surely will be others. Perhaps going all in on shape generation but then providing a convenient mechanism for controlling what parts to keep and what to discard is the way forward (`TypeShapeKind.None` partly achieves this, but it's an all-or-nothing mechanism). Not saying we need to do this now, but we might want to look at evolving this before a 1.0 release.

eiriktsarpalis: > if (openGenericName == "ListOfTimeOnly")

I still believe that using strings as a proxy for type equality is fundamentally problematic, particularly in the case of runtime reflection. This approach provides no possibility of detecting false positives either.

Are there any altnernative solutions to ths problem? Playing with metadata tokens perhaps? If not, I would be inclined to just use regular type equality for now.

AArnott: > DynamicallyAccessedMemberTypes.PublicParameterlessConstructor is specific to the converters scenario, but other scenarios might require other activation or access patterns.

What setting would you suggest? PublicConstructors?

> Wouldn't it suffice for the attribute to be present on the generic type definition itself?

I don't _think_ so. Attributing the return `Type` is what allows calling `Activator.CreateInstance(Type)` without an AOT safety warning being emitted.

>  Making the feature only keep the default constructor but discarding everything else is motivated by the use case at hand, but there surely will be others.

Agreed, that seems likely.  We had also talked about adding an `AssociatedShapes` property if and when a whole shape is needed.

> Not saying we need to do this now, but we might want to look at evolving this before a 1.0 release.

I appreciate the flexibility, as I'd very much like to get this merged soon so that we can start building on it in StreamJsonRpc to see what else falls out as important to get done.

> I still believe that using strings as a proxy for type equality is fundamentally problematic, particularly in the case of runtime reflection.

Runtime reflection doesn't use these strings. It's only source gen that does. Don't be put off by my pseudo-code string. The real one will be much closer to `System.Collections.Generic.List<System.TimeOnly>`. And ya, that leaves room for collisions. But as I stated on another thread, for the odds of a full name collision to be significant, we'd need more than one associated type (and maybe many). The more associated types you have, the more likely they'll come from multiple assemblies. Multiple assemblies in fact is a requirement to worry about type name collisions. But before you trip up on type name collisions (which is a statistically remote scenario) you will almost certainly trigger unwanted assembly loads.

> Are there any altnernative solutions to ths problem? Playing with metadata tokens perhaps? If not, I would be inclined to just use regular type equality for now.

I've asked Copilot and researched this myself. I cannot find any alternatives that don't involve a method that loads types just to do the type equivalence check. I'm certain that assembly loads will be noticed and pushed back _hard_ on. So please, I'd really rather design this to avoid them in the first place. 
I've done the metadata token thing with MEF. It's infamously unstable. In this case there wouldn't even be a cache to invalidate. Not even .NET assemblies themselves reference each others' types by metadata token because they tend to change simply from a recompile (even without source changes). 
I even considered type checks within per-assembly methods, and we test assembly identity of the requested type and then branch to another method that tests types within that assembly. That _would_ avoid unwanted assembly loads while giving us type checks. The problem is it would break on type forwarding across assemblies.

If it helps... the public API itself deals strictly with types. So if we identify a scenario in the future where strings are misbehaving, and we can't solve it, we can always switch to Type-based source-gen instead of strings. It can be done in a backward compatible way I believe by offering a source-gen option to control it, and at runtime the library can notice whether the Type or String-based delegate is initialized for a previously compiled assembly and invoke the appropriate one.

AArnott:
Switch to returning Type instead of `object` complete.

docs/shape-providers.md
AArnott: If it's conflicts that you're primarily concerned about, C# compiler [will break compilation](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYYAqAngA4Cm6AsgBRRwAM6AzgJT6HF9QB2DhwDuASwAuAYwAWPJH0VL0AIgCCK9AF4AfOglVqAewBmTVpwA0vZcvWbd+w6YZiAdhKs3FAXwDc3j7IPkA===) with an error in that case as it is. Sooner or later, we could adjust the source generator to detect conflicts before they are emitted and switch to the Type-style function. 
Of course another risk is that there isn't a _conflict_, but the input Type has the same string representation as one this function handles, though the types are not in fact equivalent. This could cause an `InvalidCastException` in the program when it tries to cast the result. We couldn't predict this in the source generator. But if the bug is raised from a user of this library, we can add a Type-based function and either switch to it or make that a separate entrypoint (so folks can pick between accuracy and avoiding assembly loads).

src/PolyType.Roslyn/KnownSymbols.cs
```diff
@@ -19,6 +24,12 @@ public class KnownSymbols(Compilation compilation)
     public IAssemblySymbol CoreLibAssembly => _CoreLibAssembly ??= Compilation.GetSpecialType(SpecialType.System_Int32).ContainingAssembly;
     private IAssemblySymbol? _CoreLibAssembly;
 
+    /// <summary>
+    /// Gets the DynamicallyAccessedMembersAttribute, if present.
+    /// </summary>
+    public INamedTypeSymbol? DynamicallyAccessedMembersAttribute => GetOrResolveType("System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute", ref _DynamicallyAccessedMembersAttribute);
```
AArnott, line 30: Lots of monsters lurking here...

`GetTypeByMetadataName` (which all these members are initialized with) is problematic in general because it will return null when more than one match is found in the compilation and/or all reference assemblies. It's better to look in the specific assembly you expect it in, but that can also be vulnerable to types being moved. 
It also returns a type even if it's `internal` in another assembly and therefore inaccessible.

The `DynamicallyAccessedMembers` attribute that is required on the local function can only be defined when that attribute exists. I implemented this, but it omits the attribute on netstandard2.0 targeted assemblies, which *may* be used in a .NET 8 context where AOT is applied. 
Polyfills such as PolySharp will declare the attribute with `internal` visibility in the local assembly. An FSharp assembly defines this attribute as `internal`, so I have to verify that the attribute's visibility is `public` before emitting it. But that precludes the possibility of a polyfill defined attribute in the local compilation, which is important for netstandard2.0 compatibility with AOT.

Emitting the attribute ourselves when it's missing creates collisions when a polyfill source generator like PolySharp is also going to emit it. 

So... I'm thinking we should rewrite this test to be more careful (I'll lean on what I've done in CsWin32 for inspiration) about matching. Then we should either emit the attribute ourselves when it's missing or emit a warning when it's not present. What do you think?

AArnott: I pushed another commit so that the attribute is always applied, and it is emitted as a polyfill when necessary.
Users who _also_ use PolySharp or some other source generator-based polyfill package may see a type collision in this case, and this can always be resolved by the user simply copying the generated attribute into their checked in source code (so both generators can see it and suppress generation), or in the case of PolySharp, the user can set an msbuild property to suppress generation of that particular type.

samples/CSharpSamples/AssociatedTypes.cs
eiriktsarpalis: Oh, ok. Would it make sense to move to a different folder, preferably under docs/?

docs/shape-providers.md
eiriktsarpalis: > > DynamicallyAccessedMemberTypes.PublicParameterlessConstructor is specific to the converters scenario, but other scenarios might require other activation or access patterns.
> 
> What setting would you suggest? PublicConstructors?

My original suggestion to use `Type` instead of `Func<object>` was because I was hoping we could avoid hardcoding the ad-hoc parameterless constructor requirement. In that sense, replacing `Func<object>` with `DynamicallyAccessedMembers.PublicParameterlessConstructor` doesn't make a whole lot of difference from my perspective. If anything, `Func<object>` is more explicit about what this feature provides.

> > Wouldn't it suffice for the attribute to be present on the generic type definition itself?
> 
> I don't _think_ so. Attributing the return `Type` is what allows calling `Activator.CreateInstance(Type)` without an AOT safety warning being emitted.

True, but it would be a false positive warning in this case.

AArnott: > True, but it would be a false positive warning in this case.

I don't understand why it's a false positive. In my experimentation, the source generated code that actually returns the Type _must_ carry the `[return: dynamicmembers]` attribute or you get warnings and at runtime an AOT app _does in fact fail_. But I didn't try applying the attribute "on the generic type definition itself". What would that look like?

>  If anything, `Func<object>` is more explicit about what this feature provides.

True. Are you thinking it's better if I revert the change I made from your last suggestion then?

samples/CSharpSamples/AssociatedTypes.cs
AArnott: Sure. I'll move it.

docs/shape-providers.md
AArnott: FWIW though, we can always expand the dynamic members attribute scope to expand use cases as demanded by users, and that can be done without a bunch of rework and new APIs, if we return types instead of a `Func<object>`.

eiriktsarpalis: > What would that look like?

Something like this:

```C#
[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
class MyPoco<T>();
```

So associated types for which the contract demands a parameterless constructor would have to specify it. No good way of enforcing this at compile time though, other than writing bespoke analyzers in the consuming libraries.

> > If anything, `Func<object>` is more explicit about what this feature provides.
> 
> True. Are you thinking it's better if I revert the change I made from your last suggestion then?

Up to you. I fully expect this will require a bit more iteration in the future.

AArnott: I tried moving the attribute to the generic type in a tiny AOT prototype app:

```cs
[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
class MyPoco<T>;

Type t = GetSomeType(typeof(MyPoco<>));
object? o = Activator.CreateInstance(t);
Console.WriteLine(o is not null);

static Type? GetSomeType(Type type)
{
	if (type == typeof(MyPoco<>))
	{
		return typeof(MyPoco<int>);
	}

	return null;
}
```

It crashed with no parameterless constructor. 

But moving the attribute to the return value on the `GetSomeType()` method worked, as I had tested before. 

Given you're open to refining this in the future (thank you), can we move forward with merging?

AArnott:
Build break fixed. Feel free to squash. :)

src/PolyType.Roslyn/KnownSymbols.cs
eiriktsarpalis: I agree that this is going to create issues with PolySharp. I had similar problems with other polyfilled attributes in the past, and ultimately the only good solution was to avoid depending on them altogether in generated code. I think we should just revert to your original approach using `Func<object>`.

src/PolyType/ReflectionProvider/ReflectionTypeShape.cs
eiriktsarpalis: My only qualm with the current approach is that in reflection every type is an associated type. Ideally shapes coming from the two providers should be observationally equivalent.

docs/shape-providers.md
eiriktsarpalis: Filed an issue: https://github.com/dotnet/runtime/issues/113650

src/PolyType.Roslyn/KnownSymbols.cs
AArnott: Ok. Working on it...

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.Object.cs
```diff
@@ -39,6 +41,53 @@ private void FormatObjectTypeShapeFactory(SourceWriter writer, string methodName
...
 
+    private static string FormatAssociatedTypesCloser(ObjectShapeModel objectShapeModel)
+    {
+        if (objectShapeModel.AssociatedTypes.Length == 0)
+        {
+            return "null";
+        }
+
+        SourceWriter writer = new() { Indentation = 0 };
+        writer.WriteLine(/* lang=c#-test */ """
```
eiriktsarpalis, line 52: What incantation is this?

src/PolyType.Roslyn/KnownSymbols.cs
AArnott: Shall we keep the newer restriction that it's only good for fetching _generic_ types? Earlier with `Func<object>` you made the case that it might as well work for non-generics as well, but when we switched to returning `Type` objects, you called out it was only useful for generics so I restricted it.

src/PolyType/AssociatedTypeAttributeAttribute.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace PolyType;
+
+/// <summary>
+/// An attribute to apply to other attribute classes that accept one or more <see cref="Type"/> arguments
+/// that should be considered associated types to whatever type the other attribute is applied to.
+/// </summary>
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
```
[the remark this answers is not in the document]
AArnott: For dead code, I sure have it plumbed through a few places including tests.

But ya, I think ultimately it's a remnant of my API design before we merged this into `TypeShapeAttribute`. I'll remove it.

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.Object.cs
eiriktsarpalis: Oh awesome! Should we just put the attribute on the method itself? Or does it not work well when we're emitting partial code fragments?

src/PolyType/ReflectionProvider/ReflectionTypeShape.cs
eiriktsarpalis: Ah right, because of the unreliability of querying all assemblies in the current domain?

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.Object.cs
AArnott: It works quite well for fragments too in my experience. I'll be happy to apply that change.

src/PolyType.Roslyn/KnownSymbols.cs
eiriktsarpalis: If this feature is a facility for returning factories for associated types, it should work with closed/non-generic types. If it's a safe proxy for `MakeGenericType` it should only accept generic type definitions. We pivoted back to the former, so let's make it more permissive for now.

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.Object.cs
AArnott: FYI it has no effect on `$"` interpolated strings, which it appears _most_ of your callers are.

src/PolyType/AssociatedTypeAttributeAttribute.cs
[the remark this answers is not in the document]
AArnott: Oh! No, it's not dead code. It's the attribute that closes #106 by providing me with a way to make `[MessagePackConverter(typeof(MyConverter<>))]` just work. 
This is the attribute that tells your source generator to read the Type out of my own 3rd party attribute and treat it as an associated type.

src/PolyType.Roslyn/KnownSymbols.cs
AArnott: Great. That's the way it happened to fall out when I git-reverted it. So we're good.

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.Object.cs
AArnott: `SourceWriter` is awesome. I've already copied it into two other projects I maintain.

src/PolyType/AssociatedTypeAttributeAttribute.cs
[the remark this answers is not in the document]
eiriktsarpalis: Ah right, the attribute annotation for third-party attributes. I don't see this being defined in PolyTypeKnownSymbols and Codecov complains about it being dead code which prompted my earlier remark. How is this being detected by our infrastructure?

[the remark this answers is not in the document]
AArnott: image

src/PolyType/ReflectionProvider/ReflectionTypeShape.cs
AArnott: Right. Also, "all" assemblies isn't the same set of assemblies that an assembly was compiled with. So you could end up with a larger or smaller set of assemblies, leading to non-deterministic results.

eiriktsarpalis (approved):
Thanks. Let me know once you're done pushing changes and I can merge this.

AArnott:
I'm already done. Let me know if you were expecting another change that I missed.

PR #131, opened 2025-03-13
Remove the ITypeShapeVisitor interface

### LINKED ISSUE #130: Consider dropping `ITypeShapeVisitor`
The [`ITypeShapeVisitor`](https://eiriktsarpalis.github.io/PolyType/api/PolyType.Abstractions.ITypeShapeVisitor.html) interface has grown a few times. Adding members to interfaces on .NET Framework is a binary and source breaking change. 

As per our conversation, maybe we should delete this interface and leave its abstract base class counterpart [`TypeShapeVisitor`](https://eiriktsarpalis.github.io/PolyType/api/PolyType.Abstractions.TypeShapeVisitor.html) around as part of the API stabilization effort.

AArnott:
The `TypeShapeVisitor` abstract class remains as the base for visitor implementations.
By dropping the interface, the API can evolve a little more freely because adding `virtual` members to an abstract class is not a binary breaking change like adding a member to an interface would be.

Closes #130


src/PolyType/Abstractions/TypeShapeVisitor.cs
```diff
@@ -13,7 +18,7 @@ public abstract class TypeShapeVisitor : ITypeShapeVisitor
     /// <param name="state">Defines user-provided state.</param>
     /// <returns>The result produced by the visitor.</returns>
     public virtual object? VisitObject<T>(IObjectTypeShape<T> objectShape, object? state = null)
-        => ThrowNotImplementedException(nameof(VisitObject));
+        => ThrowNotImplementedException();
```
eiriktsarpalis, line 21: Would it make sense to add test coverage for the default implementations now, given that this type is moving to the center of the abstractions?

eiriktsarpalis (approved):
Nice and clean. Thanks!

src/PolyType/Abstractions/TypeShapeVisitor.cs
AArnott: Sure. I can do that.

AArnott:
Tests added.

eiriktsarpalis:
Pushing as v0.30.1

PR #142, opened 2025-03-28
Produce reference assemblies for all projects

drewnoakes:
This improves incremental build compilation within the solution, for `netstandard2.0` projects/configurations in particular. It's enabled by default from `net5.0` onwards, but is worth turning on for earlier targets.

I don't believe it impacts any packaged artifacts. It should only speed up builds within this repo.


eiriktsarpalis (approved):
Thanks

PR #143, opened 2025-03-28
Add collection construction options

### LINKED ISSUE #33: Sets and maps with hash collision resistance
The means that these type shapes offer to construct a map or a set do not offer a way to provide a collision resistant hash function (in the form of a custom `IEqualityComparer<TKey>`. This is important when deserializing untrusted data.

drewnoakes:
Fixes #33 by allowing libraries that consume PolyType to provide `IEqualityComparer<T>` instances to be used when constructing various collection types, such as `HashSet<>` and `Dictionary<,>`.


src/PolyType.Examples/CborSerializer/CborSerializer.Builder.cs
```diff
@@ -89,20 +89,20 @@ public CborConverter<T> GetOrAddConverter<T>(ITypeShape<T> typeShape) =>
                     new CborMutableEnumerableConverter<TEnumerable, TElement>(
                         elementConverter,
                         getEnumerable,
-                        enumerableShape.GetDefaultConstructor(),
+                        enumerableShape.GetDefaultConstructor(CollectionConstructionOptions<TElement>.Default),
```
eiriktsarpalis, line 92: Could this be an optional parameter instead?

drewnoakes: I'd thought that making it nullable would involve an extra branch, but I see now that it's only during the visit pass, not during the execution of whatever the visit pass produces. I assume (benchmark level) perf is more critical for the second category of code than the first.

AArnott:
I'm going to pick this up next, since @drewnoakes is busy with other work. I can't push to his branch, so I'll push to my own and open a new PR when ready.

PR #144, opened 2025-03-28
Skip code coverage of test projects themselves

AArnott:
Two issues:

1. This doesn't work.
2. If it did work, it would remove all coverage of source generated code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ test: build
 		--results-directory $(ARTIFACT_PATH)/testResults \
 		--collect "Code Coverage;Format=cobertura" \
 		--logger "trx" \
-		 --settings test.runsettings \
+		--settings test.runsettings \
 		-- \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -28,6 +28,7 @@ test: build
 		--results-directory $(ARTIFACT_PATH)/testResults \
 		--collect "Code Coverage;Format=cobertura" \
 		--logger "trx" \
+		 --settings test.runsettings \
```
eiriktsarpalis, line 31:
```suggestion
		--settings test.runsettings \
```

eiriktsarpalis:
Why did you end up closing this? I had forgotten to hit merge but otherwise it looks good to me.

AArnott:
Because per the description, it doesn't work. When I ran the tests manually, I can see that code coverage still reports on the code that was intended to be excluded.

PR #150, opened 2025-04-02
Fix source gen reporting of required-by-attribute properties

AArnott:
This gets non-default source-generation constructor paths to be used in the absence of any `required` or `init` properties, when there _are_ properties annotated with `PropertyShapeAttribute.IsRequired` = `true`.

### DIFF
--- a/tests/PolyType.SourceGenerator.UnitTests/CompilationTests.cs
+++ b/tests/PolyType.SourceGenerator.UnitTests/CompilationTests.cs
@@ -267,6 +267,25 @@ public static class CompilationTests
     }
 
     [Fact]
+    public static void IsRequiredProperty()
+    {
+        Compilation compilation = CompilationHelpers.CreateCompilation("""
+            using PolyType;
+            #nullable enable
+
+            [GenerateShape]
+            public partial class Foo
+            {
+                [PropertyShape(IsRequired = true)]
+                public string? PropA { get; set; }
+            }
+            """);
+
+        PolyTypeSourceGeneratorResult result = CompilationHelpers.RunPolyTypeSourceGenerator(compilation);
+        Assert.Empty(result.Diagnostics);
+    }
+
+    [Fact]
     public static void MultiplePartialContextDeclarations_NoErrors()
     {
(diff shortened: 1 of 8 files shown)


tests/PolyType.SourceGenerator.UnitTests/CompilationTests.cs
```diff
@@ -266,6 +266,25 @@ public partial record DerivedClassWithShadowingMember : BaseClassWithShadowingMe
...
+    public static void IsRequiredProperty()
+    {
+        Compilation compilation = CompilationHelpers.CreateCompilation("""
+            using PolyType;
+            #nullable enable
+
+            [GenerateShape]
+            public partial class cFoo
+            {
+                [PropertyShape(IsRequired = true)]
```
eiriktsarpalis, line 279: Do we have a test with a `required` property set to `IsRequired = false`?

tests/PolyType.SourceGenerator.UnitTests/CompilationTests.cs
```diff
@@ -266,6 +266,25 @@ public partial record DerivedClassWithShadowingMember : BaseClassWithShadowingMe
...
 
+    [Fact]
+    public static void IsRequiredProperty()
+    {
+        Compilation compilation = CompilationHelpers.CreateCompilation("""
+            using PolyType;
+            #nullable enable
+
+            [GenerateShape]
+            public partial class cFoo
```
eiriktsarpalis, line 277:
```suggestion
            public partial class Foo
```

tests/PolyType.SourceGenerator.UnitTests/CompilationTests.cs
AArnott: Yes:

https://github.com/eiriktsarpalis/PolyType/blob/b7a7008bb1eed90dba5cc1d8c78e08f0d08d84fb/src/PolyType.TestCases/TestTypes.cs?plain=1#L2289C1-L2295C1

eiriktsarpalis (approved):
Thanks

Issue #140, opened 2025-03-27
Add expose an `AreRequiredParameters` predicate to `IConstructorShape`


eiriktsarpalis:
Validating whether required parameters have been set for a particular constructor shape currently requires a bit of added state on user defined converters. The `TArgumentState` type already encapsulates a bit set tracking which parameters have been set so that optional parameters are only set conditionally. We could piggypack on that state so that users can also check if all required parameters have been set:

```C#
public partial interface IConstructorShape<TDeclaringType, TArgumentSTate>
{
    public Func<TArgumentState, bool> GetAreRequiredParametersSet();
}
```

This would let users run validations after `TArgumentState` has been populated using data from a payload:

```C#
if (!areRequiredParametersSet(ref argumentState))
{
    throw new Exception("Some bespoke message about required parameters not being set");
}

return parameterizedCtor(ref argumentState);
```

cc @AArnott

AArnott:
I'm starting work on this.

AArnott:
Let me know if you want a PR with the content of https://github.com/AArnott/Nerdbank.MessagePack/blob/f4f1e6f8dff53f167adea321cdbc0c18a1b923a6/src/Nerdbank.MessagePack/Converters/PropertyAssignmentTrackingManager%601.cs (see https://github.com/AArnott/Nerdbank.MessagePack/pull/373) for a whole example of how to use it.

eiriktsarpalis:
Given that we're planning on incorporating this functionality in the argument state eventually, perhaps it makes sense to hold off from incorporating a separate helper for now. We can revisit if we decide that actually the argument state shouldn't have this responsibility.

AArnott:
With #176, this can potentially be done post-1.0, given we gain the ability to add features without binary breaking changes, assuming we're OK with `null` coming back from this method when it isn't supported by source generated shape providers.

PR #151, opened 2025-04-03
Add `IConstructorShape.GetAreRequiredParametersSet()` method

AArnott:
This is just the skeletal structure of the solution so far.

#140

### DIFF
--- a/src/PolyType/Abstractions/Delegates.cs
+++ b/src/PolyType/Abstractions/Delegates.cs
@@ -30,6 +30,15 @@ public delegate void Setter<TDeclaringType, TPropertyType>(ref TDeclaringType ob
 public delegate TDeclaringType Constructor<TArgumentState, TDeclaringType>(ref TArgumentState state);
 
 /// <summary>
+/// Delegate representing a function that accepts a single input taken by reference, and produces a single output.
+/// </summary>
+/// <typeparam name="TInput">The input type to the function.</typeparam>
+/// <typeparam name="TResult">The output type of the function.</typeparam>
+/// <param name="input">The function input.</param>
+/// <returns>The function output.</returns>
+public delegate TResult InFunc<TInput, TResult>(in TInput input);
+
+/// <summary>
 /// Delegate representing a constructor that accepts a span of values.
 /// </summary>
 /// <typeparam name="T">The element type of the span parameters.</typeparam>
--- a/tests/PolyType.Tests/JsonSpecializedTests.cs
+++ b/tests/PolyType.Tests/JsonSpecializedTests.cs
@@ -1,0 +1,155 @@
+﻿using PolyType.Examples.JsonSerializer;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace PolyType.Tests;
+
+public abstract partial class JsonSpecializedTests(ProviderUnderTest providerUnderTest)
+{
+    [Theory, PairwiseData]
+    public void RequiredMembersWithDefaultCtor_Exhaustive([CombinatorialRange(0, (1 << RequiredMembersWithDefaultCtor.FieldCount) - 1)] int mask)
+    {
+        StringBuilder builder = new("{");
+        if ((mask & 0x1) == 0x1) builder.Append("\"A\":\"X\",");
+        if ((mask & 0x2) == 0x2) builder.Append("\"B\":3,");
+        if ((mask & 0x4) == 0x4) builder.Append("\"C\":false,");
+        if ((mask & 0x8) == 0x8) builder.Append("\"D\":false,");
+        if (builder.Length > 1) builder.Length -= 1; // Remove trailing comma
+        builder.Append('}');
+        TestContext.Current.TestOutputHelper?.WriteLine(builder.ToString());
+
+        if ((mask & 0x5) == 0x5)
+        {
+            RequiredMembersWithDefaultCtor? actual = AssertRequirementMet<RequiredMembersWithDefaultCtor>(builder.ToString());
+            Assert.True(actual is { A: "X", C: false });
+            Assert.Equal((mask & 0x2) == 0x2 ? 3 : 5, actual.B);
+            Assert.Equal((mask & 0x8) == 0x8 ? false : true, actual.D);
+        }
+        else
+        {
+            AssertRequirementsNotMet<RequiredMembersWithDefaultCtor>(builder.ToString());
+        }
+    }
+
+    [Theory, PairwiseData]
+    public void RequiredMembersWithComprehensiveCtor_Exhaustive([CombinatorialRange(0, (1 << RequiredMembersWithComprehensiveCtor.FieldCount) - 1)] int mask)
+    {
+        StringBuilder builder = new("{");
+        if ((mask & 0x1) == 0x1) builder.Append("\"A\":\"X\",");
+        if ((mask & 0x2) == 0x2) builder.Append("\"B\":3,");
+        if ((mask & 0x4) == 0x4) builder.Append("\"C\":false,");
+        if ((mask & 0x8) == 0x8) builder.Append("\"D\":false,");
+        if (builder.Length > 1) builder.Length -= 1; // Remove trailing comma
+        builder.Append('}');
+        TestContext.Current.TestOutputHelper?.WriteLine(builder.ToString());
+
+        if ((mask & 0x5) == 0x5)
+        {
+            RequiredMembersWithComprehensiveCtor? actual = AssertRequirementMet<RequiredMembersWithComprehensiveCtor>(builder.ToString());
+            Assert.True(actual is { A: "X", C: false });
+            Assert.Equal((mask & 0x2) == 0x2 ? 3 : 5, actual.B);
+            Assert.Equal((mask & 0x8) == 0x8 ? false : true, actual.D);
+        }
+        else
+        {
+            AssertRequirementsNotMet<RequiredMembersWithComprehensiveCtor>(builder.ToString());
+        }
+    }
+
+    [Theory, PairwiseData]
+    public void RequiredMembersWithPartialCtor_Exhaustive([CombinatorialRange(0, (1 << RequiredMembersWithPartialCtor.FieldCount) - 1)] int mask)
+    {
+        StringBuilder builder = new("{");
+        if ((mask & 0x1) == 0x1) builder.Append("\"A\":\"X\",");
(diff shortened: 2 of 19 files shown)


src/PolyType/Abstractions/Delegates.cs
```diff
@@ -29,6 +29,15 @@ namespace PolyType.Abstractions;
...
 public delegate TDeclaringType Constructor<TArgumentState, TDeclaringType>(ref TArgumentState state);
 
+/// <summary>
+/// Delegate representing a function that accepts a single input taken by reference, and produces a single output.
+/// </summary>
+/// <typeparam name="TInput">The input type to the function.</typeparam>
+/// <typeparam name="TResult">The output type of the function.</typeparam>
+/// <param name="input">The function input.</param>
+/// <returns>The function output.</returns>
+public delegate TResult InFunc<TInput, TResult>(in TInput input);
```
eiriktsarpalis, line 39: Given that `TInput` is typically a mutable struct, should this be `ref readonly` to eliminate the possibility of defensive copies?

AArnott: Oh, that's subtle. I haven't used `ref readonly` much. Is the defensive copy that you're thinking of done in the case that `TInput` defines a non-readonly method or property accessor? And would `ref readonly` prevent such method calls?

eiriktsarpalis: I believe that's the main use case. In the case of argument state types in constructors it's almost always a value tuple which exposes mutable fields. For the sake of simplicity might even make sense to just make it `ref` since it's always being accessed opaquely

AArnott: I asked grok for the difference between `in` and `ref readonly` and its only good answer was around using `ref readonly` for return values -- not parameters.
I also found via sharplab that `ref readonly` doesn't avoid the defensive copy: [`in`](https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCiDetE8QSwDsAXAUwCcCBDAG2QBZEBZACkMQGVEBnASk21yCuAOgBiAe3HMeAbgE4AvrCUxYXImQCuAYyIdMiefmLkqtEAwlS+GRCpVA=) vs [`ref readonly`](https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCiDetE8QSwDsAXAUwCcCBDAG2QBZEBZACjJIDNE3KATAewLUAnogDKiAM4BKTNlzyJAOgBifPsykBuOTgC+sfTFgSiZAK4BjImMyId+YuSq0QDVepkZEhw0A=). 

The only ways to avoid the defensive copy is to use `ref` or add `readonly` to the method on the struct.

I tested ValueType [here](https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCiDetE8QSwDsAXAUwCcCBDAG2QBZEBZACkMVeIBpEAjAez7UAlIgDOIrDFzTkEAJzNRAOgCSpALYQhAbmy4AvrEMxYoomQCuAYyKIAypkR6chUhRr1EAMQHMJiY2MgA==) and no defensive copy is made. So I'm inclined to leave it as `in`.

AArnott:
The data structures kept by the argument state, constructor, and property setters are tricky, and the nuances of constructor required vs. policy required values are such that my confidence is rather low that I can make these changes in the time I have available at least for now. I'll leave this PR active as a draft I guess, and either I'll pick it up eventually or someone else can.

AArnott:
I'm making good progress on the sourcegen provider support for this. After that, I _hope_ I'll understand the area well enough to more quickly apply it to the reflection providers.

tests/PolyType.Tests/JsonSpecializedTests.cs
```diff
@@ -0,0 +1,155 @@
...
+    public partial record RequiredMembersWithPartialCtor
+    {
+        internal const int FieldCount = 4;
+
+        public RequiredMembersWithPartialCtor(string a)
+        {
+            A = a;
+        }
+
+        public string A { get; set; } // TODO: required keyword here causes the JsonSerializer to throw.
```
AArnott, line 143: @eiriktsarpalis you might be interested in this. I think this is a pre-existing bug in your sample json serializer. 
Your visitor class throws when I add `required` to this property.

AArnott:
I'm shifting focus to the reflection providers now.

AArnott:
One design question is whether tuples should consider all their members to be required or not.

AArnott:
It's done. All tests pass. Please review carefully and let me know if you can think of test holes.

AArnott:
Another design question: usability is poor for discovering _which_ required property is missing. Right now all we get is a `bool` result for the overall object. It would be far preferable if one (or all) of the missing required properties were named in the exception that I throw based on a `false` result. That would require more code generation. We could probably keep it optimized where we test for a successful match in bulk (where we do now), and only review each property individually in failure cases to improve the error.
But we still only return `bool`, so that would have to change. 
Alternatively, alongside the parameter value setter on the shape we could add an `IsSpecified` getter so that upon getting a `false` result, the deserializer could loop over all the parameters and determine which are both required and unset. That may be the best, but it will mean code generation to track all parameters regardless of whether any are required for a completely implemented API. At the moment, I only add the property flags for tracking required properties when any properties are actually required.

AArnott:
I have a branch of Nerdbank.MessagePack where I've adopted this change. All tests pass. So, there's a bit more coverage from that.

eiriktsarpalis:
> Another design question: usability is poor for discovering _which_ required property is missing. Right now all we get is a `bool` result for the overall object. It would be far preferable if one (or all) of the missing required properties were named in the exception that I throw based on a `false` result. That would require more code generation. We could probably keep it optimized where we test for a successful match in bulk (where we do now), and only review each property individually in failure cases to improve the error.
> But we still only return `bool`, so that would have to change.

That's a great point ,reporting back on the missing members is essential for any production-grade serializer. At the same time, there is a related feature of [having deserializers fail if duplicate properties are encountered](https://github.com/dotnet/runtime/issues/108521). We're building this in STJ for .NET 10 because it's an SDL requirement. I wonder if there is a design that addresses both concerns.

AArnott:
Merge conflicts resolved

AArnott:
An interface for the arg struct would result in _boxing_ the arg struct if folks ever used the interface. That seems undesirable, especially since this would be used in success paths.

> One possibility is changing the signature of the delegate to be `InFunc<TArgumentState, int[]> GetUnsetRequiredParameters()` which returns the indices of the unset parameters. 

True. That would improve the use case for this PR and its backing issue.
However, it wouldn't help solve the duplicate assignment case.

Adding `IsPropertySet`/`IsParameterSet` methods would address serve _both_ purposes though. It would allow the serializer library to avoid double sets on its own and throw instead (when the policy for that is desirable), and it would allow the library to figure out which required property is missing.

eiriktsarpalis:
> An interface for the arg struct would result in _boxing_ the arg struct if folks ever used the interface. That seems undesirable, especially since this would be used in success paths.

Not unless you explicitly upcast, i.e. instead of
```C#
TArgumentState state = ...; // Some opaque generic state type
IArgumentState boxedState = state;
_ = boxedState[i].AssignmentState;
```
doing
```C#
TArgumentState state = ...; // Some opaque generic state type
state[i].AssignmentState;
```
would ensure that the JIT generates bespoke devirtualized machine code for the given generic type.

AArnott:
So a generic type constraint would be applied to the TArgState everywhere it appeared to guarantee to the compiler that the type implements that interface then. Makes sense.

Are you thinking this interface might also expose a `bool RequiredParametersSet {get;}` property on it?

eiriktsarpalis:
> Are you thinking this interface might also expose a `bool RequiredParametersSet {get;}` property on it?

Oh I do like that!

AArnott:
These ideas around tracking assignments in argument state don't help us track assignments for objects that don't use arg state at all. That's fine for my use case for this PR, but for tracking double-assignments, it puts the onus of tracking squarely on the deserializer. And if the deserializer has to _sometimes_ track which properties it has seen, I wonder how much benefit there really is to saving the deserializer the trouble for the arg state case.

AArnott:
Given I need to track default ctor property setting separately for the double-assignment, I whipped that up in Nerdbank.MessagePack tonight to see how hard it would be, and how natural it would be to also track required parameter setting. Pretty natural, actually. So while there might be a _small_ perf win perhaps by having the source generated property setter do the work with a hard-coded bit flag, I'm leaning on abandoning this PR and keeping what I have built. It's not quite done, but I think another hour or two will do it.

Incidentally, what I built is entirely uncoupled from the rest of my library. So if you want to, I can replace this PR with a contribution of the class that tracks property and parameter assignments so that all PolyType serialization libraries can use it.

eiriktsarpalis:
> @eiriktsarpalis did you notice my question at [#151 (comment)](https://github.com/eiriktsarpalis/PolyType/pull/151#issuecomment-2832562952)

I did not, but yeah I agree that tuples should mark all their elements as required since that better mirrors how the language handles them.

> Incidentally, what I built is entirely uncoupled from the rest of my library. So if you want to, I can replace this PR with a contribution of the class that tracks property and parameter assignments so that all PolyType serialization libraries can use it.

Up to you. Out of curiosity, does it allocate at all for constructors with a small number of parameters?

AArnott:
> does it allocate at all for constructors with a small number of parameters?

No. And it recycles BitArray objects that are required for over 64 parameters/properties.

eiriktsarpalis:
Ah, so it implements an object pool presumably?

AArnott:
yes

AArnott:
Check out https://github.com/AArnott/Nerdbank.MessagePack/pull/373/files#diff-8d1c5ada3a1b800b871006a8df40e5e8dbd8ede60611c76680f21fb8ed7efc99 for the new class that handles both required parameter and double assignment checks.

Let me know if you think something close to that would be a good fit for PolyType.

AArnott:
>> @eiriktsarpalis did you notice my question at https://github.com/eiriktsarpalis/PolyType/pull/151#issuecomment-2832562952

> I did not, but yeah I agree that tuples should mark all their elements as required since that better mirrors how the language handles them.

Well, now that I look at the shapes for `ValueTuple` generic types, you give them a default ctor and read/write properties (which ... matches the definition). But that means that their properties are not naturally required.
I suppose if they are to be considered required, PolyType will have to start marking their properties as such.

eiriktsarpalis:
> Well, now that I look at the shapes for `ValueTuple` generic types, you give them a default ctor and read/write properties (which ... matches the definition). But that means that their properties are not naturally required. I suppose if they are to be considered required, PolyType will have to start marking their properties as such

That's surprising. I have no recollection of why I did it that way but I imagine it's because it makes for simpler codegen in the ref emit provider when dealing with nested tuple types. I'll file an issue.

PR #152, opened 2025-04-06
Rename diagnostic code prefix from TS to PT

### LINKED ISSUE #129: Diagnostic codes are reminiscent of TypeShape
Your diagnostic codes start with TS (for TypeShape) instead of PT or some other prefix.

AArnott:
Fixes #129


eiriktsarpalis (approved):
Thanks

PR #154, opened 2025-04-08
Fix bug in source generation around generic ctor parameter with default value

AArnott:
Fixes a code gen bug where generic type parameters that are set to `default` are assigned `null` by generated code, even when the generic type argument is a value type.

```
        51:                 ArgumentStateConstructorFunc = static () => null!,
        70:                 DefaultValue = null!,
```

### DIFF
--- a/src/PolyType.Roslyn/Helpers/RoslynHelpers.cs
+++ b/src/PolyType.Roslyn/Helpers/RoslynHelpers.cs
@@ -464,7 +464,7 @@ internal static class RoslynHelpers
 
         string literalExpr = parameter.ExplicitDefaultValue switch
         {
-            null => "null!",
+            null => parameter.Type.IsNullable() ? "null!" : "default",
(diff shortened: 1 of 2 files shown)


eiriktsarpalis:
The shape providers use tuples to represent argument states, with the exception of single-parameter constructors that use the type of the parameter itself. What I suspect is happening here is that the generator is correctly inferring that the default value is `null` instead of `default`.

AArnott:
> I take it you hit this while working on https://github.com/eiriktsarpalis/PolyType/pull/151?

No. I hit it while writing a test for #155. It took about an hour to realize that the fault wasn't mine, and exactly what about the test caused the compilation to fail. Then it was easy to avoid in my test, so I wrote a new test to keep track of it and put it in this separate branch.

eiriktsarpalis (approved):
Thanks!

PR #155, opened 2025-04-08
Replace associated type factories with more flexible associated *shapes*

### LINKED ISSUE #146: AssociatedShapes now required
In adapting StreamJsonRpc to use PolyType and Nerdbank.MessagePack, I've encountered a case where I have a _data type_ that is generic and must be closed over a generic type argument found in a shapeable object.

Use of `IAsyncEnumerable<int>` means not only that we need a constructor for `AsyncEnumerableConverter<T>` (which associated _types_ gives us) but also `EnumerationResults<int>`, which must be serializable and thus we need a shape for it.

Should we add an `AssociatedShapes` property alongside the `AssociatedTypes` property for `TypeShapeExtensionAttribute` and `TypeShapeAttribute`?

AArnott:
In addition to associated types that may need to be constructed (e.g. generic serialization converters), we may also need whole associated _shapes_. In the case of StreamJsonRpc, an associated _type_ (a converter) makes use of yet another _shape_ to aid in conversion of part of the value. But as the same generic type argument has to run through all these types, we need an associated _shape_ to be generated as well:

```mermaid
flowchart LR
  A["IAsyncEnumerable&lt;int><br>(natural shape)"] --> B["AsyncEnumerableConverter&lt;int><br>(Associated type with constructor only)"]
  B --> C["EnumerationResult&lt;int><br>(Whole shape required by converter)"]
```

Attribute construction has very limited syntax, so I've removed the `AssociatedTypes` property from `TypeShapeAttribute` (which does not allow multiple attributes for a single type) and created a new `AssociatedTypeShapeAttribute` class with `AllowMultiple = true`. Each attribute occurrence can specify the 'depth' of the required shape and one or more types for which that depth is required. This way, many types with varying depth requirements may be associated with some other type.

The existing `TypeShapeExtensionAttribute` required only an added depth property because it already allowed multiple occurrences.

Internally, we aggregate all attribute sources for associated types into a single definition of an associated type, with flags that indicate all the shape elements that are required.

The acceptance test for associated types has been updated as well. The case may not look particularly compelling, so in short the case is that an associated *type* may be a custom converter and an associated *shape* may be a convertible type that the custom converter uses as a part of its serialization.

- [x] Docs updated
 
Closes #146


src/PolyType.SourceGenerator/Model/EquatableEnum.cs
```diff
@@ -0,0 +1,13 @@
...
+
+/// <summary>
+/// An equatable wrapper around an enum.
+/// </summary>
+/// <param name="Value">The enum vlaue</param>
+/// <remarks>
+/// This is useful for use with <see cref="ImmutableEquatableDictionary{TKey, TValue}" /> which requires its type arguments to
+/// implement <see cref="IEquatable{T}" /> (and enums don't, but are nevertheless equatable).
+/// </remarks>
+public record struct EquatableEnum<T>(T Value) where T : Enum;
```
eiriktsarpalis, line 13: Ugh. I kind of assumed this was being extended to enums.

src/PolyType/Abstractions/AssociatedTypeRequirements.cs
```diff
@@ -0,0 +1,21 @@
...
+/// Describes the requirements for preparing an associated type.
+/// </summary>
+[Flags]
+public enum AssociatedTypeRequirements
+{
+    /// <summary>No requirements.</summary>
+    None = 0x0,
+
+    /// <summary>
+    /// A factory method should be prepared for the associated type.
```
eiriktsarpalis, line 13: What alternative name to "factory" could we use to further emphasize that this specifically pertains to default constructors?

src/PolyType/Abstractions/ITypeShape.cs
```diff
@@ -60,6 +60,24 @@ public interface ITypeShape
...
+    /// If the associated type is a generic type definition, the type arguments used on this shape's <see cref="Type"/>
+    /// will be used to close the associated generic type.
+    /// </param>
+    /// <returns>The shape for the associated type, or <see langword="null" /> if no shape for the associated type is available.</returns>
+    /// <remarks>
+    /// <see cref="ReflectionProvider.ReflectionTypeShapeProvider"/> can produce the shape on demand without any <see cref="TypeShapeAttribute.AssociatedShapes"/>,
+    /// while <see cref="SourceGenModel.SourceGenTypeShapeProvider"/> is expected to only produce the shape that was explicitly requested via attribute.
+    /// </remarks>
+    /// <exception cref="InvalidOperationException">Thrown when this method is called on an instance that does not represent a generic type.</exception>
+    ITypeShape? GetAssociatedTypeShape(Type associatedType);
```
eiriktsarpalis, line 80: I feel that this method could be unified with the existing `GetAssociatedTypeFactory`. `ITypeShape` already exposes default constructors where available (although it takes a bit of digging with visitors to extract it from the cases that support it).

Would it make sense to expose a `Func<object> GetDefaultConstructor();` method on the `ITypeShape` object directly? This could either be implemented by the interface directly or exposed as an extension method that applies a visitor. In that case `AssociatedTypeRequirements` would morph into an enum that controls the degree/depth of shape generation for a given type: e.g. `DefaultConstructorOnly`, `Transitive`, or possible other modes we can come up with in the future.

src/PolyType.SourceGenerator/Model/EquatableEnum.cs
AArnott: I was surprised by this as well.

src/PolyType/Abstractions/ITypeShape.cs
AArnott: I can see that.
But I wonder how much more complicated it could potentially make visitors if they have to account for partially complete shapes.

AArnott: I'm sketching this out over here: https://github.com/AArnott/PolyType/pull/1
Your feedback is appreciated

tests/PolyType.Tests/AssociatedTypesTests.cs
```diff
@@ -49,17 +55,25 @@ public void CanConstructAssociatedType_ByExtension()
...
+        IObjectTypeShape? associatedShape = (IObjectTypeShape?)typeShape.GetAssociatedTypeShape(typeof(GenericDataTypeVerifier<,>));
+        Assert.NotNull(associatedShape);
+        Func<object>? factory = associatedShape.GetDefaultConstructor();
         Assert.NotNull(factory);
         Assert.IsType<GenericDataTypeVerifier<int, string>>(factory.Invoke());
+
+        if (partialShapesSupported)
+        {
+            // Verify the associated type's shape is only partially available.
+            Assert.Empty(associatedShape.Properties);
```
AArnott, line 67: Should querying for properties on a partial shape throw or merely return an empty collection? I'm thinking throwing may be more appropriate to avoid confusion when collections are empty when they otherwise wouldn't be. But if we throw, we should probably add a depth property to the shape so that a program that cares may avoid an exception while exploring arbitrary shapes.

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
```diff
@@ -8,4 +8,33 @@ namespace PolyType.Roslyn;
...
+/// Keep this in sync with the TypeShapeDepth enum defined in the PolyType assembly.
+/// </devremarks>
+[Flags]
+public enum TypeShapeDepth
+{
+    /// <summary>No shape is required.</summary>
+    None = 0x0,
+
+    /// <summary>
+    /// A constructor should be included in the shape, if one is declared.
```
eiriktsarpalis, line 27: Would this also include parameterized constructors and its parameter types?

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
```diff
@@ -8,4 +8,33 @@ namespace PolyType.Roslyn;
...
+
+    /// <summary>
+    /// Properties should be included in the shape, if any are declared.
+    /// </summary>
+    Properties = 0x2,
+
+    /// <summary>
+    /// The shape should be fully generated.
+    /// </summary>
+    All = Constructor | Properties,
```
eiriktsarpalis, line 39:
```suggestion
    Full = Constructor | Properties,
```

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
```diff
@@ -8,4 +8,33 @@ namespace PolyType.Roslyn;
...
+    /// <summary>No shape is required.</summary>
+    None = 0x0,
+
+    /// <summary>
+    /// A constructor should be included in the shape, if one is declared.
+    /// </summary>
+    Constructor = 0x1,
+
+    /// <summary>
+    /// Properties should be included in the shape, if any are declared.
```
eiriktsarpalis, line 32:
```suggestion
    /// Properties and their types should be included in the shape, if any are declared.
```

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
eiriktsarpalis: I presume from the definition of `All` that the above is true, therefore:
```suggestion
    /// A constructor , its parameters and their types should be included in the shape, if one is declared.
```

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
```diff
@@ -8,4 +8,33 @@ namespace PolyType.Roslyn;
...
+/// <summary>
+/// Describes the requirements for preparing an associated type.
+/// </summary>
+/// <devremarks>
+/// Keep this in sync with the TypeShapeDepth enum defined in the PolyType assembly.
+/// </devremarks>
+[Flags]
+public enum TypeShapeDepth
+{
+    /// <summary>No shape is required.</summary>
```
eiriktsarpalis, line 23: I presume this would still result in an empty shape `ITypeShape<T>` for the type being generated, correct?
```suggestion
    /// <summary>Generates an empty <see cref="ITypeShape{T}" /> without any transitive shape information.</summary>
```

src/PolyType/Abstractions/ShapeExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+/// </summary>
+public static class ShapeExtensions
+{
+    /// <summary>
+    /// Retrieves the default constructor for the given type shape, if available.
+    /// </summary>
+    /// <typeparam name="T">The type to be constructed.</typeparam>
+    /// <param name="shape">The type's object shape.</param>
+    /// <returns>A delegate for the default constructor; or <see langword="null" /> if none is available.</returns>
+    public static Func<T>? GetDefaultConstructor<T>(this IObjectTypeShape<T> shape) => (Func<T>?)shape.Constructor?.Accept(DefaultConstructorVisitor.Instance);
```
eiriktsarpalis, line 16: This can further be extended to collection shapes that have a `Mutable` construction strategy. Perhaps though it just makes sense that we expose this method directly on `ITypeShape<T>` paired with a `HasDefaultConstructor` property?

src/PolyType/TypeShapeAttribute.cs
```diff
@@ -45,21 +45,5 @@ public TypeShapeKind Kind
...
-    /// <para>
-    /// Each type must declare a public default constructor.
-    /// </para>
-    /// <para>
-    /// If the type this attribute is applied to is a generic type definition,
-    /// each type in this array must also be a generic type definition
-    /// with the same number of generic type parameters.
-    /// </para>
-    /// </remarks>
-    public Type[] AssociatedTypes { get; init; } = [];
```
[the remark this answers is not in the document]
eiriktsarpalis: Sounds good

tests/PolyType.Tests/AssociatedTypesTests.cs
eiriktsarpalis: Good question. I think throwing is appropriate here (particularly if the reflection-based typeshape of the same type enumerates the properties correctly).

tests/PolyType.Tests/AssociatedTypesTests.cs
```diff
@@ -1,19 +1,25 @@
 ﻿using PolyType;
 using PolyType.Tests;
+using Xunit.Internal;
 
-[assembly: TypeShapeExtension(typeof(AssociatedTypesTests.GenericDataType<,>), AssociatedTypes = [typeof(AssociatedTypesTests.GenericDataTypeVerifier<,>)])]
+[assembly: TypeShapeExtension(typeof(AssociatedTypesTests.GenericDataType<,>), AssociatedShapeDepth = TypeShapeDepth.Constructor, AssociatedTypes = [typeof(AssociatedTypesTests.GenericDataTypeVerifier<,>)])]
+[assembly: TypeShapeExtension(typeof(AssociatedTypesTests.GenericDataType<,>), AssociatedShapeDepth = TypeShapeDepth.All, AssociatedTypes = [typeof(AssociatedTypesTests.ExtraShape<,>)])]
```
eiriktsarpalis, line 6: Do we have any tests checking that a depth is overriddent to `All` if the type is explicilty specified either directly via `GenerateShape` or transitively because a type containing it is generated?

eiriktsarpalis (commented):
I must say I love the direction this feature is taking, thanks for taking the time to work on it.

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
```diff
@@ -8,4 +8,33 @@ namespace PolyType.Roslyn;
...
+public record struct AssociatedTypeModel(INamedTypeSymbol AssociatedType, IAssemblySymbol AssociatingAssembly, Location? Location, TypeShapeDepth Requirements);
+
+/// <summary>
+/// Describes the requirements for preparing an associated type.
+/// </summary>
+/// <devremarks>
+/// Keep this in sync with the TypeShapeDepth enum defined in the PolyType assembly.
+/// </devremarks>
+[Flags]
+public enum TypeShapeDepth
```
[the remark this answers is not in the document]
AArnott: In some places I use the term 'requirements' instead of 'depth'. That's slightly shorter than `TypeShapeGenerationMode`. How about I rename to `TypeShapeRequirements` then?

src/PolyType/Abstractions/ShapeExtensions.cs
AArnott: Generic `IEnumerableTypeShape` and `IDictionaryTypeShape` interfaces already define `GetDefaultConstructor()` methods. Implementing this extension method on `ITypeShape` or `ITypeShape<T>` seems to be impossible because those interfaces do not offer a `Constructor` property.

`ITypeShape<T>` being an interface, it cannot be implemented there (while supporting .NET Standard 2.0). And since the method body is the same for all interface implementations, an extension method is the most natural place to put it IMO.
IMO the bool property that predicts whether this method will return `null` doesn't add enough value to add more public API. But I'll go with your suggestion if you insist.

tests/PolyType.Tests/AssociatedTypesTests.cs
AArnott: I thought so. But now I can't find it. I'll add one.

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
AArnott: Honestly I didn't anticipate anyone using this. It's just customary to define a 0 value on flags enums. I suppose I should test its effect though.

AArnott: The behavior _was_ that we wouldn't generate empty shapes. But based on your expectation, I've modified the code and added tests to verify empty shapes are generated.

AArnott (commented):
All comments responded to. Most suggestions applied. Let me know what you think.

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
eiriktsarpalis: Sounds good. An empty shape is still useful since it can be used with an `ITypeShapeFunc` for example to extract results from a `ValueTask<T>`.

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
```diff
@@ -8,4 +8,33 @@ namespace PolyType.Roslyn;
...
+
+    /// <summary>
+    /// A constructor, its parameters and their types should be included in the shape, if one is declared.
+    /// </summary>
+    Constructor = 0x1,
+
+    /// <summary>
+    /// Properties and their types should be included in the shape, if any are declared.
+    /// </summary>
+    Properties = 0x2,
```
eiriktsarpalis, line 34: It just occurred to me, but what would be the behavior of either `Constructor` or `Properties` when used with collection or dictionary types? My expectation is that the two should be identical, but does this suggest that we need to pick more generic sounding names for the values?

Similarly, how does this behave when applied to union types? (particularly when one of the union cases is a collection?)

tests/PolyType.SourceGenerator.UnitTests/NumberedLineWriter.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. Full rights reserved.
```
eiriktsarpalis, line 1: Seems this got subbed by accident
```suggestion
﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
AArnott: The feature only supports object types. I couldn't readily define how this might be used in any other case.

AArnott: I don't have (and at the moment, can't imagine) a use case for any other type than object either. So... if you can think of any change I should make to communicate this through the API, I'd prefer that to expanding scope at this point. I can mention it in the API and concept docs.

AArnott:
Docs updated.

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
eiriktsarpalis: How does it behave like at the moment? I would presume `None` similarly produces empty shapes, `Full` includes the key/value/element shapes and `Constructor` and `Properties` are no-ops unless targeting union types?

src/PolyType/Abstractions/AssociatedTypeAttributeAttribute.cs
```diff
@@ -14,7 +14,11 @@ public class AssociatedTypeAttributeAttribute : Attribute
     /// The name of the attribute constructor parameter or named argument (i.e. attribute property)
     /// whose <see cref="Type"/> or <see cref="Type"/> array argument should be included as associated types.
     /// </param>
-    /// <param name="requirements">The requirements to apply to each associated type.</param>
+    /// <param name="requirements">
+    /// The requirements to apply to each associated type.
+    /// Only applicable to object shapes.
+    /// All other shapes (e.g. collections, enums, unions) are always completely defined.
```
eiriktsarpalis, line 20: I think this answers my earlier question. I could possibly nitpick the `None` vs. `Full` being applicable to all types but I can't see any real use-case right now. Union types are a bit nuanced since one might expect object hierarchies to inherit the annotation from the base, but I'm fine with leaving this underspecified for now.

src/PolyType/AssociatedTypeShapeAttribute.cs
```diff
@@ -19,6 +19,10 @@ public class AssociatedTypeShapeAttribute(params Type[] associatedTypes) : Attri
     /// Gets the elements of the generated shape that are required at runtime.
     /// </summary>
     /// <value>The default value is <see cref="TypeShapeRequirements.Full"/>.</value>
+    /// <remarks>
+    /// This property only impacts generation of object shapes.
+    /// All other shapes (e.g. collections, enums, unions) are always completely defined.
```
eiriktsarpalis, line 24: Do we have test coverage for non-object types?

docs/shape-providers.md
```diff
@@ -167,25 +167,39 @@ public record MyPoco<T>(T Value)
 #### Associated types
 
 The `TypeShape` attribute can also be used to specify associated types for the target type.
-This provides a Reflection-free way to activate an associated type via its default constructor
-given its @PolyType.Abstractions.ITypeShape.
+This provides a Reflection-free way to leap from one @PolyType.Abstractions.ITypeShape to an associated shape, and get all the functionality of the associated shape.
```
eiriktsarpalis, line 170:
```suggestion
In the context of source generated/AOT applications, this provides a Reflection-free way to leap from one @PolyType.Abstractions.ITypeShape to an associated shape, and get all the functionality of the associated shape.
```

docs/shape-providers.md
```diff
@@ -167,25 +167,39 @@ public record MyPoco<T>(T Value)
...
 For example, serializers may need to jump from a type shape to its converter.
 
 [!code-csharp[](CSharpSamples/AssociatedTypes.cs#TypeShapeOneType)]
 
-@PolyType.Abstractions.ITypeShape.GetAssociatedType*?displayProperty=nameWithType is the method to use to obtain a factory for instances of the associated type.
+Use the @PolyType.Abstractions.ITypeShape.GetAssociatedTypeShape*?displayProperty=nameWithType method to obtain the shape of an associated type.
 The @PolyType.SourceGenModel.SourceGenTypeShapeProvider implementation of this method requires that the associated types be pre-determined at compile time via attributes.
 The @PolyType.ReflectionProvider.ReflectionTypeShapeProvider does _not_ require these attributes.
```
eiriktsarpalis, line 177:
```suggestion
The @PolyType.ReflectionProvider.ReflectionTypeShapeProvider is guaranteed to work with all types and therefore does _not_ require these attributes.
```

eiriktsarpalis (approved):
Thanks

src/PolyType.Roslyn/Model/AssociatedTypeModel.cs
AArnott: In my update yesterday I added API docs to describe current behavior, which is that the enum is ignored for anything except object shapes. 

There's a method in the source gen parser with a massive switch statement (effectively) for each type of shape, and nearly all of them just force `Full` rather than propagate the specified flags. This is because I haven't implemented anything else for each of these other shapes. Collections, enums, tuples,... don't have properties, and not having constructors would break them entirely, so it didn't make sense to me to implement partial shapes.

AArnott:
I pushed fixes for the doc comments. Let me know if you think the testing of non-object associated types is a blocking issue.

eiriktsarpalis:
Published as v0.35.1

PR #158, opened 2025-04-20
Add support for types implementing `IAsyncEnumerable<T>`

eiriktsarpalis:
Extends the `IEnumerableTypeShape` type to include support for types implementing `IAsyncEnumerable<T>`. Note that this is a breaking change cc @AArnott


AArnott:
I'll check this out later today. Very interesting.

AArnott (commented):
I'm curious what use case led to this new feature.
Async enumeration seems to stray a bit into the *behavior* world rather than the data world since async enumeration is generally exposed as return types of methods rather than properties. And they're typically backed by data someplace other than process memory.

No objection to the change I guess. I'm curious how this will interplay with StreamJsonRpc's IAsyncEnumerable support which is highly specialized.

PR #160, opened 2025-04-28
Add `TypeShapeExtensionAttribute.Marshaller` property

### LINKED ISSUE #118: Reverse surrogate types?
I just encountered [a scenario with @dotfede](https://github.com/MessagePack-CSharp/MessagePack-CSharp/issues/2129#issuecomment-2670480532) that is interesting. He wants to serialize a type in another library, which he owns but it cannot take a dependency on an assembly for attributes. No problem. PolyType and [Nerdbank.MessagePack](https://github.com/AArnott/Nerdbank.MessagePack) can handle that by default.
But he also wants to get serialization behavior that would require attributes. And he doesn't want to write a custom formatter to achieve it. That sounds like what surrogates would be great at. Except that the pattern today requires the original type to be annotated to point to the surrogate.

So what if we add a reverse-direction for surrogates? That sounds generally very useful. Surrogates address a need when the original type shape isn't ideal and we want to define a different one. That may be quite often encountered when you can't attribute the original type.
So in @dotfede's scenario, while Project A cannot have any attributes but _does_ define some of the types to be serialized, Project B could define a surrogate type and use attributes somehow so that when Project B generates type shapes for Project A (since Project A doesn't offer them for its own types) it will generate them based on the surrogate instead.

What do you think, @eiriktsarpalis?

AArnott:
This effectively adds 'reverse surrogates', where instead of a type declaring its own surrogate (which requires PolyType awareness), a consuming assembly can prescribe a surrogate for an external type.

* [x] Add tests
* [x] Define a new diagnostic code for reporting when type shape extensions contain conflicting parameters.
* [x] Support in source generator type shape provider.
* [x] Support in reflection-based type shape providers.
* [x] Formalize the process of merging type shape extensions across assemblies.
        For the source generator, the compilation assembly has the last word on merge conflicts.
        For reflection based shape providers, the assembly list employs a 'first one wins' policy.
        We report a new analyzer diagnostic if one assembly defines conflicting type shape extensions within itself.

Closes #118 

## In other news

* I bumped the StyleCop.Analyzers version because the one you were on before was from 2019 and didn't know what a `record struct` was, so it was reporting incorrect diagnostics for the docs on them. This led me to make a few syntax touch-ups in unrelated files. The more significant changes I'm "saving for later" by suppressing the warnings via .editorconfig.
* I introduced the first proper analyzer, and the first analyzer tests that leverage's @sharwell's roslyn analyzer test harness, which I hope you'll agree is pretty great, considering how easy it is to write a test and pinpoint the location of any expected diagnostics.
* We should be working toward moving more diagnostics from the source generator to analyzers. This is the guidance from the roslyn team and provides a better IDE experience for the user due to more responsive updates to reported diagnostics (since analyzers run far more frequently than source generators).

### DIFF
--- a/src/PolyType.Roslyn/IncrementalTypes/EquatableDiagnostic.cs
+++ b/src/PolyType.Roslyn/IncrementalTypes/EquatableDiagnostic.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.CodeAnalysis;
 using PolyType.Roslyn.Helpers;
+using System.Collections.Immutable;
 
 namespace PolyType.Roslyn;
 
@@ -30,7 +31,7 @@ public readonly struct EquatableDiagnostic(
     /// <summary>
     /// Additional locations for the diagnostic.
     /// </summary>
-    public IEnumerable<Location>? AdditionalLocations { get; init; }
+    public Location[] AdditionalLocations { get; init; } = [];
 
     /// <summary>
     /// Creates a new <see cref="Diagnostic"/> instance from the current instance.
@@ -46,7 +47,8 @@ public readonly struct EquatableDiagnostic(
     {
         return Descriptor.Equals(other.Descriptor) &&
             MessageArgs.SequenceEqual(other.MessageArgs) &&
-            Location == other.Location;
+            Location == other.Location &&
+            AdditionalLocations.SequenceEqual(other.AdditionalLocations);
     }
 
     /// <inheritdoc/>
@@ -56,6 +58,11 @@ public readonly struct EquatableDiagnostic(
         foreach (object? messageArg in MessageArgs)
         {
             hashCode = CommonHelpers.CombineHashCodes(hashCode, messageArg?.GetHashCode() ?? 0);
+        }
+
+        foreach (Location location in AdditionalLocations)
+        {
+            hashCode = CommonHelpers.CombineHashCodes(hashCode, location.GetHashCode());
         }
 
         hashCode = CommonHelpers.CombineHashCodes(hashCode, Location?.GetHashCode() ?? 0);
--- a/src/PolyType.Roslyn/IncrementalTypes/EquatableDiagnostic.cs
+++ b/src/PolyType.Roslyn/IncrementalTypes/EquatableDiagnostic.cs
@@ -28,10 +28,15 @@ public readonly struct EquatableDiagnostic(
     public Location? Location { get; } = location?.GetLocationTrimmed();
 
     /// <summary>
+    /// Additional locations for the diagnostic.
+    /// </summary>
+    public IEnumerable<Location>? AdditionalLocations { get; init; }
+
+    /// <summary>
     /// Creates a new <see cref="Diagnostic"/> instance from the current instance.
     /// </summary>
     public Diagnostic CreateDiagnostic()
-        => Diagnostic.Create(Descriptor, Location, MessageArgs);
+        => Diagnostic.Create(Descriptor, Location, AdditionalLocations, MessageArgs);
 
     /// <inheritdoc/>
     public override readonly bool Equals(object? obj) => obj is EquatableDiagnostic info && Equals(info);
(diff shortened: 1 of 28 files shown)


AArnott:
This is mostly ready for review. It's just the policy of merging type extensions defined across assemblies that I am working on.

src/PolyType.Roslyn/IncrementalTypes/EquatableDiagnostic.cs
```diff
@@ -27,11 +27,16 @@ public readonly struct EquatableDiagnostic(
     /// </summary>
     public Location? Location { get; } = location?.GetLocationTrimmed();
 
+    /// <summary>
+    /// Additional locations for the diagnostic.
+    /// </summary>
+    public IEnumerable<Location>? AdditionalLocations { get; init; }
```
eiriktsarpalis, line 33: Needs to be one of the special collection types to avoid reference equality:
```suggestion
    public ImmutableEquatableArray<Location>? AdditionalLocations { get; init; }
```

AArnott: `Location` does not satisfy the type constraint of `IEquatable<Location>`. But I'll update your Equals/GetHashCode methods here to make up for not using an equatable array type.

src/PolyType.Roslyn/IncrementalTypes/EquatableDiagnostic.cs
```diff
@@ -27,11 +28,15 @@ public readonly struct EquatableDiagnostic(
     /// </summary>
     public Location? Location { get; } = location?.GetLocationTrimmed();
 
+    /// <summary>
+    /// Additional locations for the diagnostic.
+    /// </summary>
+    public Location[] AdditionalLocations { get; init; } = [];
```
eiriktsarpalis, line 34:
```suggestion
    public Location[] AdditionalLocations { get; init; } = [];

```

src/PolyType.SourceGenerator/Model/TypeExtensionModel.cs
```diff
@@ -17,5 +17,15 @@ public sealed record TypeExtensionModel
...
+
+    /// <summary>
+    /// Gets an optional marshaller type that can be used to convert the <see cref="Target"/> type to a surrogate type.
+    /// </summary>
+    public INamedTypeSymbol? Marshaller { get; init; }
+
+    /// <summary>
+    /// Gets the locations of the <see cref="TypeShapeExtensionAttribute"/> attributes that target the <see cref="Target"/> type.
+    /// </summary>
+    public ImmutableArray<Location> Locations { get; init; } = ImmutableArray<Location>.Empty;
```
eiriktsarpalis, line 30: Similarly, this will result in false positive equality comparisons in the incremental model. If `Location` doesn't implement IEquatality perhaps a wrapper struct is in order?

src/PolyType.SourceGenerator/TypeShapeExtensionAnalyzer.cs
```diff
@@ -0,0 +1,71 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+using System.Collections.Immutable;
+
+namespace PolyType.SourceGenerator;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class TypeShapeExtensionAnalyzer : DiagnosticAnalyzer
```
eiriktsarpalis, line 9: Nit: Maybe nest in an `Analyzers` folder?

AArnott:
I dropped `net472` as a target framework for the source generator unit test project on macOS and linux. It is now failing there (probably due to the roslyn analyzer test harness) because net472 on these OSs is emulated by mono, which doesn't support something the new test harness is doing. But no one _builds_ with mono any more anyway (even if they _run_ on it), so skipping tests on mono that emulate _build_ behavior seems appropriate.

eiriktsarpalis (approved):
Thanks

PR #164, opened 2025-05-06
Normalize line endings

AArnott:
Add a .gitattributes file to set a firm policy for how git should record line endings, ensuring that you don't get PR diffs showing lines changed that only changed in line endings.

I forced git to reevaluate every file in the repo. Only two files ended up needing to change due to inconsistent line endings compared to the rest of the repo.


eiriktsarpalis (approved):
Thanks

Issue #148, opened 2025-04-01
Recognize dictionary expression factory methods


eiriktsarpalis:
.NET 10 is likely going to add support for dictionary expressions. We should make sure that PolyType recognizes whatever factory shapes do get shipped via https://github.com/dotnet/runtime/issues/114090.

eiriktsarpalis:
cc @drewnoakes @AArnott. The fact that the proposed factories do accept `IEqualityComparer<TKey>` is affirming our direction to also include that in PT.

PR #165, opened 2025-05-09
Allow providing comparers to keyed collections

### LINKED ISSUE #33: Sets and maps with hash collision resistance

### LINKED ISSUE #167: Shape providers inconsistently support internal constructors on collection shapes
The source gen shape provider supports instantiating collections using internal constructors. Reflection shape providers require public constructors.

Object type shapes consistently support only public constructors.

It seems then that the 'fix' should be that the sourcegen shape provider should require public constructors.

AArnott:
Changes the enumerable and dictionary shapes to allow construction with an `IEqualityComparer<T>` or `IComparer<T>` when the underlying collection types support it (and follow one of the expected API conventions).

Closes #33
Closes #143 (by superseding it)
Fixes #167

Progress in work involved:

* [x] Design API changes
* [x] Add a property to the shapes that indicates whether the comparer will be used.
* [x] A decent amount of tests
* [x] Exhaustive tests
* [x] Source generator type shape provider changes
* [x] Non-emit Reflection type shape provider changes
* [x] Emit reflection type shape provider changes
* [x] Doc update


src/PolyType.Roslyn/Model/EnumerableDataModel.cs
```diff
@@ -41,6 +42,12 @@ public sealed class EnumerableDataModel : TypeDataModel
     /// </summary>
     public required IMethodSymbol? FactoryMethod { get; init; }
 
+    /// <summary>
+    /// Constructor or static factory method whose shape is governed by <see cref="ConstructionStrategy"/>
+    /// and declares either an <see cref="IEqualityComparer{T}"/> or <see cref="IComparer{T}"/> parameter.
+    /// </summary>
+    public required IMethodSymbol? FactoryMethodWithComparer { get; init; }
```
eiriktsarpalis, line 49: Does this need to be a separate field? It seems one could just look at the arguments of `FactoryMethod` to determine if it accepts a comparer.

src/PolyType.Roslyn/ModelGenerator/TypeDataModelGenerator.Dictionary.cs
```diff
@@ -46,81 +47,95 @@ private bool TryMapDictionary(ITypeSymbol type, ImmutableArray<AssociatedTypeMod
...
+        if (factoryMethod is null &&
+            namedType.Constructors.FirstOrDefault(ctor
+                => IsAccessibleSymbol(ctor)
+                && ctor is { Parameters: [{ Type: { } parameterType }] }
+                && IsSpanOfKeyValuePair(parameterType, keyType, valueType)) is IMethodSymbol ctor2)
         {
             constructionStrategy = CollectionModelConstructionStrategy.Span;
             factoryMethod = ctor2;
+
+            // .ctor(ReadOnlySpan<KeyValuePair<K, V>>, I[Equality]Comparer<T>) or .ctor(I[Equality]Comparer<T>, ReadOnlySpan<KeyValuePair<K, V>>)
```
eiriktsarpalis, line 74: Unless I've already done this, we should add an issue tracking support for the final shape of dictionary collection builders: https://github.com/dotnet/runtime/issues/114090

src/PolyType.SourceGenerator/Model/ConstructionWithComparer.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// The construction expects just the <see cref="IEqualityComparer{T}"/>.
+    /// </summary>
+    EqualityComparer,
+
+    /// <summary>
+    /// The construction expects an <see cref="IEqualityComparer{T}"/> then the values.
+    /// </summary>
+    EqualityComparerValues,
+
+    /// <summary>
```
eiriktsarpalis, line 35: Something I was discussing with @drewnoakes is we could avoid a combinatorial explosion of the various parameter permutations by modelling argument as an array of enums, i.e.
```C#
[IEqualityComparerOfT]
[IEnumerableOfT, IComparerOfT],
[SpanOfT, IEqualityComparerOfT]
```
etc.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -74,9 +74,10 @@ public interface IDictionaryTypeShape<TDictionary, TKey, TValue> : ITypeShape<TD
     /// <summary>
     /// Creates a delegate wrapping a parameterless constructor of a mutable collection.
     /// </summary>
+    /// <param name="collectionConstructionOptions">Optional options that control how the collection is constructed.</param>
     /// <exception cref="InvalidOperationException">The collection is not <see cref="CollectionConstructionStrategy.Mutable"/>.</exception>
     /// <returns>A delegate wrapping a default constructor.</returns>
-    Func<TDictionary> GetDefaultConstructor();
+    Func<TDictionary> GetDefaultConstructor(in CollectionConstructionOptions<TKey> collectionConstructionOptions = default);
```
eiriktsarpalis, line 80: Should we consider building in capacity support to avoid future breaking changes? If it makes the diff too complicated, we could just leave it as an a no-op in the shape providers.
```suggestion
    Func<int?, TDictionary> GetDefaultConstructor(in CollectionConstructionOptions<TKey> collectionConstructionOptions = default);
```

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -88,14 +89,16 @@ public interface IDictionaryTypeShape<TDictionary, TKey, TValue> : ITypeShape<TD
     /// <summary>
     /// Creates a constructor delegate for creating a collection from a span.
     /// </summary>
+    /// <param name="collectionConstructionOptions">Optional options that control how the collection is constructed.</param>
     /// <exception cref="InvalidOperationException">The collection is not <see cref="CollectionConstructionStrategy.Span"/>.</exception>
     /// <returns>A delegate constructing a collection from a span of values.</returns>
-    SpanConstructor<KeyValuePair<TKey, TValue>, TDictionary> GetSpanConstructor();
+    SpanConstructor<KeyValuePair<TKey, TValue>, TDictionary> GetSpanConstructor(in CollectionConstructionOptions<TKey> collectionConstructionOptions = default);
```
eiriktsarpalis, line 95: Given that the options is only passed once by the visitor, I don't think making it a struct is necessary.

src/PolyType.Roslyn/Model/EnumerableDataModel.cs
AArnott: This allows for a runtime decision to call one factory method versus another based on whether or not the user provided a comparer. This supports the use case that I believe exists that some overloads that accept a comparer reject a null argument.

src/PolyType.SourceGenerator/Model/ConstructionWithComparer.cs
AArnott: Yes, he did that. But he had an array of an array of enums. It was significantly more complicated than just a simple enum and I didn't see any reason for the complexity and extra allocations because I can't imagine what else we might add to this.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
AArnott: Maybe not. But seeing as I've already done all the work and it's a struct, why redo it? It does avoid an allocation for each call to obtain a delegate after all.

src/PolyType.Roslyn/ModelGenerator/TypeDataModelGenerator.Dictionary.cs
AArnott: You already have one: https://github.com/eiriktsarpalis/PolyType/issues/148

src/PolyType/Abstractions/IDictionaryTypeShape.cs
AArnott: I think you mentioned elsewhere the (potentially over-engineered) possibility of taking a custom struct as input instead of just `int?`. If we think we _might_ support allowing comparers to be provided at activation time instead of visitor-time, a custom struct approach might defend against that future possibility. Thoughts?
I'm leaning toward saving this for a subsequent change. I like the idea of front-loading it to reduce breaking changes on the roadmap, but I'd also like to get _this_ comparer change in ASAP so that we can keep building on it within the higher-level libraries.

AArnott:
I added properties to the shapes to indicate whether comparers are supported by the constructor.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
eiriktsarpalis: > If we think we _might_ support allowing comparers to be provided at activation time instead of visitor-time, a custom struct approach might defend against that future possibility. Thoughts?

While I can't discount the possibility of this ever coming up as a requirement, it seems unlikely (payloads describing their own equality comparison? Sounds kind of dangerous). That said, maybe I'm overestimating the performance penalty incurred by the ctor delegate having a bit more branching than strictly necessary. Maybe we should just keep things simple and have one options struct accepted by the delegate.

AArnott: Regarding the conditional branching, here's another idea: what if instead of a method that returns just one delegate, it returns a struct of delegates. One delegates takes an `int` and one takes no args. That way, no branching is necessary at any level, unless a converter can't predict whether a collection will be length prefixed or not, and will have to branch anyway based on that.
By returning a struct of delegates instead of adding new ITypeShape members that return one delegate each, we allow for backward compatibility when adding new delegates.

eiriktsarpalis: In the case of the reflection provider, wouldn't this require that all delegates be generated even if they're not being used?

AArnott: Yes. But I suppose we can mitigate that by returning a _class_ of delegates instead of a struct of delegates. If the class defined virtual property getters for the delegates, the reflection providers can override the getters to generate the delegates on-demand.

eiriktsarpalis: I think ultimately this creates a shape type within a shape type. It could be exposed as a separate factory directly on shape type.

AArnott: But shapes are _interfaces_ which makes adding members a breaking change. My proposal allows adding members that _don't_ require API breaking changes.

eiriktsarpalis: Yeah.. I should get to changing those to abstracts at some point, at least the generic ones.

AArnott: > It could be exposed as a separate factory directly on shape type.

If this is the way you want to go, does that mean we'll double up like this, for each existing factory?

```cs
Func<TDictionary> GetDefaultConstructor(in CollectionConstructionOptions<TKey> collectionConstructionOptions = default);
Func<int, TDictionary> GetDefaultConstructorWithCapacity(in CollectionConstructionOptions<TKey> collectionConstructionOptions = default);
```

In any case, I'd ask that capacity factories be held off for a subsequent PR.

eiriktsarpalis: > In any case, I'd ask that capacity factories be held off for a subsequent PR.

Fine by me.

AArnott:
Tests all pass now. But I have some TODOs to fill in, and some overall review to ensure consistency.
There are also some clearly _not_ implemented methods, so evidently we have some test holes since no tests are failing.

AArnott:
It's a draft so I can work on code coverage, which is finding some small bugs. But please do review what I have.

AArnott:
I've tested this in Nerdbank.MessagePack and it works. It allows me to address the last parity feature and support secure deserialization.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -31,6 +31,11 @@ public interface IDictionaryTypeShape : ITypeShape
     /// Gets the construction strategy for the given collection.
     /// </summary>
     CollectionConstructionStrategy ConstructionStrategy { get; }
+
+    /// <summary>
+    /// Gets the kind of custom comparer (if any) that this collection may be initialized with.
+    /// </summary>
+    ComparerConstruction CustomComparerSupport { get; }
```
eiriktsarpalis, line 38:
```suggestion
    CollectionComparerOptions ComparerOptions { get; }
```

src/PolyType/ReflectionProvider/ReflectionParameterShape.cs
```diff
@@ -40,8 +41,18 @@ public ReflectionParameterShape(
...
+    public Setter<TArgumentState, TParameter> GetSetter()
+    {
+        if (_setter is { } setter)
+        {
+            return setter;
+        }
+
+        lock (_syncObject)
+        {
+            return _setter ??= _provider.MemberAccessor.CreateConstructorArgumentStateSetter<TArgumentState, TParameter>(_ctorInfo, Position);
```
eiriktsarpalis, line 53: What necessitated introduction of locks here?

tests/PolyType.Tests/PolyType.Tests.csproj
```diff
@@ -1,12 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  
+
   <PropertyGroup>
-    <TargetFrameworks>net9.0;net8.0;net472</TargetFrameworks>
+    <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">$(TargetFrameworks);net472</TargetFrameworks>
```
eiriktsarpalis, line 5: I had enabled x-plat net472 testing intentionally so that it can be tested in non-Windows dev environments and  also to make sure that mono works with polytype.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
AArnott: Can you imagine any more fields being added to this struct? I can't think of what that might be.

src/PolyType/ReflectionProvider/ReflectionParameterShape.cs
eiriktsarpalis: Makes sense. When writing such code I typically factor the critical path into a local method so that the getter method itself can more easily, i.e. `return _setter ?? CreateGetterHelper();`. I think we can also use `Interlocked.CompareExchange` here but the downside is potential duplicated allocations if there is contention at startup.

AArnott: Fine idea. I'll implement your `??` local function idea.

tests/PolyType.Tests/PolyType.Tests.csproj
AArnott: I removed it because mono didn't like the presence of the ModuleInitializer that I added. But I'll workaround it another way so you can keep testing on mono then.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
eiriktsarpalis: > Can you imagine any more fields being added to this struct?

I cannot, but that's the very definition of features being added retroactively to options types.

> Also, I only pass the struct by reference, so even if it did grow large, it shouldn't be a problem.

But enabling that necessitates introducing concepts like `InFunc`. I think it also incorrectly communicates to the user that this is a performance oriented method that should be used in the hot path.

AArnott: It doesn't require the user to make any special considerations in their code though. They don't have to use `in` (or even notice it) when calling this method, they can still omit the optional argument, and... I'm not sure what harm or confusion would come from the user noticing. I actually have been using `in` modifiers on struct parameters more frequently in ordinary code just to avoid a copy when the callee has no business changing the struct, so IMO it's a normal thing that shouldn't be avoided (especially at the cost of additional allocations) just to avoid the user thinking that he should call this method in a hot-path (I don't know what that would even mean to the user in this case).

It means Nerdbank.MessagePack will allocate another object for every hashed collection type member it encounters, even if the keyed type is the same across multiple members, since otherwise I have to store a cache of these things.

That said, this isn't that important to me so I'll go ahead and make the change.

AArnott: Change pushed.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -74,9 +79,10 @@ public interface IDictionaryTypeShape<TDictionary, TKey, TValue> : ITypeShape<TD
     /// <summary>
     /// Creates a delegate wrapping a parameterless constructor of a mutable collection.
     /// </summary>
+    /// <param name="collectionConstructionOptions">Options that control how the collection is constructed. Use <see cref="IDictionaryTypeShape.ComparerOptions" /> to predict which properties may be worth initializing.</param>
     /// <exception cref="InvalidOperationException">The collection is not <see cref="CollectionConstructionStrategy.Mutable"/>.</exception>
     /// <returns>A delegate wrapping a default constructor.</returns>
-    Func<TDictionary> GetDefaultConstructor();
+    Func<TDictionary> GetDefaultConstructor(CollectionConstructionOptions<TKey>? collectionConstructionOptions = default);
```
eiriktsarpalis, line 85:
```suggestion
    Func<TDictionary> GetDefaultConstructor(CollectionConstructionOptions<TKey>? collectionConstructionOptions = null);
```

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -88,14 +94,16 @@ public interface IDictionaryTypeShape<TDictionary, TKey, TValue> : ITypeShape<TD
     /// <summary>
     /// Creates a constructor delegate for creating a collection from a span.
     /// </summary>
+    /// <param name="collectionConstructionOptions">Options that control how the collection is constructed. Use <see cref="IDictionaryTypeShape.ComparerOptions" /> to predict which properties may be worth initializing.</param>
     /// <exception cref="InvalidOperationException">The collection is not <see cref="CollectionConstructionStrategy.Span"/>.</exception>
     /// <returns>A delegate constructing a collection from a span of values.</returns>
-    SpanConstructor<KeyValuePair<TKey, TValue>, TDictionary> GetSpanConstructor();
+    SpanConstructor<KeyValuePair<TKey, TValue>, TDictionary> GetSpanConstructor(CollectionConstructionOptions<TKey>? collectionConstructionOptions = default);
```
eiriktsarpalis, line 100:
```suggestion
    SpanConstructor<KeyValuePair<TKey, TValue>, TDictionary> GetSpanConstructor(CollectionConstructionOptions<TKey>? collectionConstructionOptions = null);
```

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -88,14 +94,16 @@ public interface IDictionaryTypeShape<TDictionary, TKey, TValue> : ITypeShape<TD
...
 
     /// <summary>
     /// Creates a constructor delegate for creating a collection from an enumerable.
     /// </summary>
+    /// <param name="collectionConstructionOptions">Optional options that control how the collection is constructed. Use <see cref="IDictionaryTypeShape.ComparerOptions" /> to predict which properties may be worth initializing.</param>
     /// <exception cref="InvalidOperationException">The collection is not <see cref="CollectionConstructionStrategy.Enumerable"/>.</exception>
     /// <returns>A delegate constructing a collection from an enumerable of values.</returns>
-    Func<IEnumerable<KeyValuePair<TKey, TValue>>, TDictionary> GetEnumerableConstructor();
+    Func<IEnumerable<KeyValuePair<TKey, TValue>>, TDictionary> GetEnumerableConstructor(CollectionConstructionOptions<TKey>? collectionConstructionOptions = default);
```
eiriktsarpalis, line 108:
```suggestion
    Func<IEnumerable<KeyValuePair<TKey, TValue>>, TDictionary> GetEnumerableConstructor(CollectionConstructionOptions<TKey>? collectionConstructionOptions = null);
```

eiriktsarpalis (approved):
Thanks!

AArnott:
All requested changes pushed.

eiriktsarpalis:
Released as 0.38.1

PR #169, opened 2025-05-18
Add population support for collection types that explicitly implement collection interfaces.

AArnott:
The source generator compilation tests fail with `ExpandoObject`, but only in DEBUG builds where nullable annotations are checked. 

> 1>D:\source\PolyType\src\PolyType.TestCases\obj\Debug\net8.0\PolyType.SourceGenerator\PolyType.SourceGenerator.PolyTypeGenerator\PolyType.SourceGenerator.ShapeProvider.ExpandoObject.g.cs(21,51,21,175): error CS8631: The type 'System.Dynamic.ExpandoObject' cannot be used as type parameter 'TDictionary' in the generic type or method 'CollectionHelpers.AsReadOnlyDictionary<TDictionary, TKey, TValue>(TDictionary)'. Nullability of type argument 'System.Dynamic.ExpandoObject' doesn't match constraint type 'System.Collections.Generic.IDictionary<string, object>'.

Adding `ExpandoObject` to `TestTypes` revealed support limitations in your samples too, whose tests now fail.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -188,3 +188,6 @@ dotnet_diagnostic.CA1510.severity = none # incompatible with multitargeting proj
 
 # SA1134: Attributes should not share line
 dotnet_diagnostic.SA1134.severity = none
+
+# CA1010: Generic interface should also be implemented
+dotnet_diagnostic.CA1010.severity = none
--- a/src/PolyType.Roslyn/KnownSymbols.cs
+++ b/src/PolyType.Roslyn/KnownSymbols.cs
@@ -135,6 +135,12 @@ public class KnownSymbols(Compilation compilation)
     private Option<INamedTypeSymbol?> _DictionaryOfTKeyTValue;
 
     /// <summary>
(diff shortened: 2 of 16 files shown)


AArnott:
The issue is `ExpandoObject` implements `IDictionary<string, object?>` and the generated code only emits `IDictionary<string ,object>` (note the lack of the nullable annotation on `object`).

eiriktsarpalis:
I've pushed a fix for the build warning. The issue with the failing tests specifically is that explicitly implemented indexers are not being taken into account when deriving the construction strategy for the type. This mirrors how collection initializers work, where it similarly is not possible to use collection initializer syntax to populate an `ExpandoObject`. We might want to consider relaxing that constraint, for types implementing IList and IDictionary.

eiriktsarpalis:
I've extended support to collections using explicit interface implementations, provided that the types themselves are reference types. Struct types are not supported for the moment since they require continual boxing and unboxing to mutate. This fixes support for types like `ExpandoObject` and `LinkedList<T>`.

PR #173, opened 2025-05-21
Fix NotImplementedException thrown for `IDictionary<K, V>` with a comparer

AArnott:
As part of this, I simplify some code and reuse more delegates to reduce GC pressure. I also noticed through an added test that sourcegen and reflection shape providers produced diverging construction strategies for IDictionary. I unified them (to the mutable strategy rather than span, since that works on non-Emit reflection and is probably faster in most or all cases).


eiriktsarpalis (commented):
Thanks!

PR #176, opened 2025-05-22
Switch type shape interfaces to abstract classes

### LINKED ISSUE #171: Declare shapes as abstract classes instead of interfaces
For an interface, any additional functionality is a binary breaking change.
As an abstract class, new members can be added with default behavior such that it needn't be breaking.

The alternative to this is to start adding versioned interfaces like `IDictionaryTypeShape2<T, K, V>` with the added members.

AArnott:
This will provide a way to add functionality in many places without it constituting a binary breaking change.

Closes #171 

- [x] Delete dead code in now unused type shape base classes.
- [x] Have all abstract classes take an `ITypeShapeProvider` (even if they don't yet use it).
- [x] Consider what other interfaces should be made into abstract classes (e.g. some of the non-generic type shape ones).
- [ ] Rename the abstract classes to drop the `I` prefix.

### DIFF
--- a/src/PolyType/Abstractions/IEnumTypeShape.cs
+++ b/src/PolyType/Abstractions/IEnumTypeShape.cs
@@ -48,7 +48,7 @@ public abstract class IEnumTypeShape<TEnum, TUnderlying>(ITypeShapeProvider prov
     /// <summary>
     /// Gets the shape of the underlying type used to represent the enum.
     /// </summary>
-    public new virtual ITypeShape<TUnderlying> UnderlyingType => Provider.Resolve<TUnderlying>();
+    public new abstract ITypeShape<TUnderlying> UnderlyingType { get; }
 
     /// <inheritdoc/>
     public override Type Type => typeof(TEnum);
--- a/src/PolyType/ReflectionProvider/ReflectionEnumTypeShape.cs
+++ b/src/PolyType/ReflectionProvider/ReflectionEnumTypeShape.cs
@@ -9,5 +9,8 @@ internal sealed class ReflectionEnumTypeShape<TEnum, TUnderlying>(ReflectionType
     where TEnum : struct, Enum
 {
     /// <inheritdoc/>
+    public override ITypeShape<TUnderlying> UnderlyingType => Provider.Resolve<TUnderlying>();
+
+    /// <inheritdoc/>
     public override ITypeShape? GetAssociatedTypeShape(Type associatedType) => null;
 }


AArnott:
@eiriktsarpalis This is just an early peek, in case you're interested.

eiriktsarpalis:
Thanks. My only concern is that we won't ever be able to convert the non-generic types into abstracts so this doesn't fully realize our goal of making abstractions evolvable. Maybe if we reconsidered how the type relationships work but I don't have clear alternative in my mind.

AArnott:
I'm actually curious whether even the non-generic interfaces can be made abstract classes except for the root `ITypeShape` and `ITypeShape<T>` interfaces. I'm going to explore that.

The test failures are due to what appears to be a different interpretation of a `TParameter?` in a class vs. an interface. When it was in an interface, it was apparently a `Nullable<TParameter>` when `TParameter` was a value type. But in a class, it evidently is interpreted as just `TParameter` for a value type. This causes `(TParameter?)null` to throw NRE when `TParameter` is `int`. I haven't figured out a workaround yet.

AArnott:
> My only concern is that we won't ever be able to convert the non-generic types into abstracts

No need to fret. I made those abstract classes too. :)

AArnott:
I think this change is about ready. If you agree with the design decisions I made (e.g. how I workaround C#'s inability to both override and declare a new member of the same name), I'll proceed to rename the new abstract classes to remove the preceding `I`. I'm holding off doing that now since it'll explode the size of the diff with noise.

src/PolyType/Abstractions/IConstructorShape.cs
```diff
@@ -29,47 +29,56 @@ public interface IConstructorShape
     /// Includes all formal parameters of the underlying constructor,
     /// as well as any member that can be specified in a member initializer expression.
     /// </remarks>
-    IReadOnlyList<IParameterShape> Parameters { get; }
+    public abstract IReadOnlyList<IParameterShape> Parameters { get; }
+
+    /// <inheritdoc cref="DeclaringType"/>
+    protected abstract IObjectTypeShape DeclaringTypeNonGeneric { get; }
```
eiriktsarpalis, line 35: Not sure I understand the purpose of this member?

src/PolyType/Abstractions/IConstructorShape.cs
```diff
@@ -29,47 +29,56 @@ public interface IConstructorShape
...
     /// Gets the shape of the declaring type for the constructor.
     /// </summary>
-    new IObjectTypeShape<TDeclaringType> DeclaringType { get; }
+    public new virtual IObjectTypeShape<TDeclaringType> DeclaringType => (IObjectTypeShape<TDeclaringType>)provider.Resolve<TDeclaringType>();
+
+    /// <inheritdoc/>
+    protected override IObjectTypeShape DeclaringTypeNonGeneric => this.DeclaringType;
+
+    /// <inheritdoc/>
+    public override object? Accept(TypeShapeVisitor visitor, object? state = null) => visitor.VisitConstructor(this, state);
```
eiriktsarpalis, line 62: Do we want to mark this sealed?
```suggestion
    public sealed override object? Accept(TypeShapeVisitor visitor, object? state = null) => visitor.VisitConstructor(this, state);
```

src/PolyType/Abstractions/IConstructorShape.cs
eiriktsarpalis: Is it because you can't declare a `new` virtual with the same name in the generic type?

src/PolyType/Abstractions/IEnumTypeShape.cs
```diff
@@ -1,26 +1,64 @@
...
 /// <typeparam name="TUnderlying">The underlying type used to represent the enum.</typeparam>
-public interface IEnumTypeShape<TEnum, TUnderlying> : ITypeShape<TEnum>, IEnumTypeShape
+public abstract class IEnumTypeShape<TEnum, TUnderlying>(ITypeShapeProvider provider) : IEnumTypeShape(provider), ITypeShape<TEnum>
     where TEnum : struct, Enum
 {
     /// <summary>
     /// Gets the shape of the underlying type used to represent the enum.
     /// </summary>
-    new ITypeShape<TUnderlying> UnderlyingType { get; }
+    public new virtual ITypeShape<TUnderlying> UnderlyingType => Provider.Resolve<TUnderlying>();
```
eiriktsarpalis, line 51: I would rather this be determined by the deriving type as opposed to hardcoding it to the provider.

src/PolyType/Abstractions/IEnumTypeShape.cs
```diff
@@ -1,26 +1,64 @@
...
+    public override Type Type => typeof(TEnum);
+
+    /// <inheritdoc/>
+    protected override ITypeShape UnderlyingTypeNonGeneric => UnderlyingType;
+
+    /// <inheritdoc/>
+    public override object? Accept(TypeShapeVisitor visitor, object? state = null) => visitor.VisitEnum(this, state);
+
+    /// <inheritdoc/>
+    public override object? Invoke(ITypeShapeFunc func, object? state = null) => func.Invoke(this, state);
```
eiriktsarpalis, line 63: I wish there was a way we could have a `TypeShape<T>` class so that this implementation wouldn't need to be replicated everywhere.

src/PolyType/Abstractions/IConstructorShape.cs
AArnott: It's because you cannot declare an `override` *and* declare a `new virtual` with the same member name.
The way I've done it though, using `protected`, observers will have exactly the same experience as before as far as completion lists, virtual/override behavior, etc. It's only the implementations that have to override this slightly unusual name.

src/PolyType/Abstractions/IConstructorShape.cs
AArnott: In general I rarely apply `sealed` to anything (class or member) and think it's overused as it ties the hands of people like myself that want to do interesting thing later. But I'm perfectly willing to add it if _you_ want me to. I doubt there's a scenario for further overriding this method, so `sealed` seems harmless in this case.

src/PolyType/Abstractions/IEnumTypeShape.cs
```diff
@@ -1,26 +1,64 @@
-﻿namespace PolyType.Abstractions;
+﻿using System.Reflection;
+
+namespace PolyType.Abstractions;
 
 /// <summary>
 /// Provides a strongly typed shape model for a .NET enum.
 /// </summary>
-public interface IEnumTypeShape : ITypeShape
+public abstract class IEnumTypeShape(ITypeShapeProvider provider) : ITypeShape
```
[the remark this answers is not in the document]
AArnott: It _could_ be. But why? It's not like there's likely to be interesting novel behavior in the get accessor.

src/PolyType/Abstractions/IEnumTypeShape.cs
AArnott: It's `virtual`. And in some cases there are multiple reflection-based implementations that all do the same thing, so defining it here seems both guaranteed to be correct and in most cases optimal. 
When defining abstract types, I prefer virtual when reasonable and correct behavior can be assumed, yet a derived type might optimize it (as the sourcegen implementation does).

If you feel strongly, sure, I can make this abstract instead. It just feels like more work now, so that future implementations will have to do more work to get off the ground.

src/PolyType/Abstractions/IEnumTypeShape.cs
AArnott: Agreed. But that creates the multi-inheritance problem. Because if `ITypeShape<T>` is a class, then our generic kind-specific classes like `EnumShape<T>` would have to derive both from `EnumShape` (non-generic) and `TypeShape<T>`, which of course C# doesn't allow.

src/PolyType/Abstractions/IEnumTypeShape.cs
eiriktsarpalis: The default implementation isn't necessary in the case of the source generator. In addition, there are certain corner cases (thinking about union case shapes) where you actually need to return a different shape for the same type. So if this only really applies to the reflection provider, we might as well be intentional and keep it abstract.

AArnott: Fixed.

src/PolyType/Abstractions/IConstructorShape.cs
AArnott: Change applied to Accept and Invoke methods everywhere.

PR #182, opened 2025-05-31
Discover associated types via attributes on members

### LINKED ISSUE #181: AssociatedTypeAttribute is ignored when an attribute that uses it is applied to *members*
`MessagePackConverterAttribute` defined in Nerdbank.MessagePack works great for associated types when the attribute is applied to a *type*. But no associated types are created when the attribute is applied to *members*.

AArnott:
Previously, we supported 3rd party custom attributes that define associated types that appeared on types. With this change, we now recognize those attributes when they appear on properties and fields.

I had to rework the model for collecting associated types since we no longer know by the time we parse a type what the whole set of its associated types are.

Fixes #181


eiriktsarpalis (approved):
Seems fairly straightforwards, thanks!

PR #183, opened 2025-06-05
Ensure all lazily-initialized properties return only one value

AArnott:
The lack of thread-safety here lead to multiple values being returned, which caused malfunctions in [one of PolyType's consumers][1].

Other occurrences of this problem already have been known to cause test instability within this repo, and those that surfaced were fixed. In this PR I went through _all_ the lazily initialized properties in all the shape implementations to ensure that I got them all.

[1]: https://github.com/AArnott/Nerdbank.MessagePack/issues/429#issuecomment-2942402813

### DIFF
--- a/src/PolyType/ReflectionProvider/ReflectionConstructorShape.cs
+++ b/src/PolyType/ReflectionProvider/ReflectionConstructorShape.cs
@@ -11,6 +11,7 @@ internal sealed class ReflectionConstructorShape<TDeclaringType, TArgumentState>
     IConstructorShapeInfo ctorInfo) :
     IConstructorShape<TDeclaringType, TArgumentState>
 {
+    private readonly object _syncObject = new();
     private IReadOnlyList<IParameterShape>? _parameters;
     private Func<TArgumentState>? _argumentStateConstructor;
     private Constructor<TArgumentState, TDeclaringType>? _parameterizedConstructor;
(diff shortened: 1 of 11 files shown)


AArnott:
FWIW, the way I've been putting locks in, is consistent with how I've done it in prior changes. But it is more complicated than I would typically default to doing it. I'd typically just put a lock around the whole property accessor. But this way avoids a lock when the field is already initialized. I don't know that it's worth the added complexity. I'm willing to do it either way. But having a common pattern that we can apply everywhere is nice, so if I change it, I think I should change it for all of them.

src/PolyType/ReflectionProvider/ReflectionConstructorShape.cs
```diff
@@ -22,7 +23,21 @@ internal sealed class ReflectionConstructorShape<TDeclaringType, TArgumentState>
...
     object? IConstructorShape.Accept(TypeShapeVisitor visitor, object? state) => visitor.VisitConstructor(this, state);
 
-    public IReadOnlyList<IParameterShape> Parameters => _parameters ??= GetParameters().AsReadOnlyList();
+    public IReadOnlyList<IParameterShape> Parameters
+    {
+        get
+        {
+            if (_parameters is null)
+            {
+                lock (_syncObject)
```
eiriktsarpalis, line 32: I would prefer if this was made lock-free using `CompareExchange`, but seeing as this pattern is being already followed elsewhere it's fine to make this change later.

PR #186, opened 2025-06-14
Add shape attribute to rename enum members

### LINKED ISSUE #72: Add `EnumShapeAttribute`
The DataContractSerializer attribute set is largely redundant with your `*ShapeAttribute` set. Except that DCS offers `EnumMemberAttribute` which allows renaming an enum value, and PolyType has no equivalent.

`IEnumTypeShape` doesn't enumerate the enum value names anyway, forcing us to use reflection at runtime. Presumably that would need to be filled in to make specifying a rename to make sense.

AArnott:
- [x] API changes for enum member shape support
- [ ] Implement in reflection shape providers
- [ ] Implement in sourcegen shape providers
- [ ] Tests
- [ ] Acceptance tests

Closes #72

### DIFF
--- a/src/PolyType/Abstractions/IEnumTypeShape.cs
+++ b/src/PolyType/Abstractions/IEnumTypeShape.cs
@@ -25,4 +25,14 @@ public interface IEnumTypeShape<TEnum, TUnderlying> : ITypeShape<TEnum>, IEnumTy
     /// Gets the shape of the underlying type used to represent the enum.
     /// </summary>
     new ITypeShape<TUnderlying> UnderlyingType { get; }
+
+    /// <summary>
+    /// Gets the names and values for each enum member.
+    /// </summary>
+    /// <devremarks>
+    /// We use <typeparamref name="TUnderlying"/> instead of <typeparamref name="TEnum"/>
+    /// to avoid needlessly closing generic types around unique value types (which swell the size of trimmed apps)
+    /// when the caller can easily cast between enum and underlying types.
+    /// </devremarks>
+    IReadOnlyDictionary<string, TUnderlying> Members { get; }
 }
--- a/src/PolyType/EnumMemberShapeAttribute.cs
+++ b/src/PolyType/EnumMemberShapeAttribute.cs
@@ -1,0 +1,13 @@
+﻿namespace PolyType;
+
+/// <summary>
+/// Annotates a member of an enum to provide additional metadata for its shape representation.
+/// </summary>
+[AttributeUsage(AttributeTargets.Field, Inherited = false, AllowMultiple = false)]
+public class EnumMemberShapeAttribute : Attribute
+{
+    /// <summary>
+    /// Gets a custom name to be used for the particular enum member in its containing <see cref="Abstractions.IEnumTypeShape"/>.
+    /// </summary>
+    public string? Name { get; init; }
+}


src/PolyType/Abstractions/IEnumTypeShape.cs
```diff
@@ -25,4 +25,14 @@ public interface IEnumTypeShape<TEnum, TUnderlying> : ITypeShape<TEnum>, IEnumTy
...
+
+    /// <summary>
+    /// Gets the names and values for each enum member.
+    /// </summary>
+    /// <devremarks>
+    /// We use <typeparamref name="TUnderlying"/> instead of <typeparamref name="TEnum"/>
+    /// to avoid needlessly closing generic types around unique value types (which swell the size of trimmed apps)
+    /// when the caller can easily cast between enum and underlying types.
+    /// </devremarks>
+    IReadOnlyDictionary<string, TUnderlying> Members { get; }
```
AArnott, line 41: I'm using `string` as the key instead of `TUnderlying` is because the name of the member is guaranteed to be unique whereas the value is _not_ guaranteed to be unique.

I'm not adding Func's that convert between the two as @eiriktsarpalis had proposed in the issue because it assumes converting between string or `ReadOnlySpan<char>` is useful, which in msgpack it will not be. And whatever _is_ useful to the PolyType user can be built up all based on this dictionary.

src/PolyType/ReflectionProvider/ReflectionEnumTypeShape.cs
```diff
@@ -1,15 +1,52 @@
...
     public override object? Accept(TypeShapeVisitor visitor, object? state = null) => visitor.VisitEnum(this, state);
     public ITypeShape<TUnderlying> UnderlyingType => Provider.GetShape<TUnderlying>();
     ITypeShape IEnumTypeShape.UnderlyingType => UnderlyingType;
+    public IReadOnlyDictionary<string, TUnderlying> Members => _members ??= InitializeMembers();
+
+    private IReadOnlyDictionary<string, TUnderlying> InitializeMembers()
+    {
+        lock (_syncObject)
+        {
+            if (_members is not null)
```
eiriktsarpalis: Shouldn't the `_members` field be set while under lock? The way this flows right now means that the result in `Members` won't be a singleton.

src/PolyType/Abstractions/IEnumTypeShape.cs
eiriktsarpalis: Is this dictionary meant to function as a substitute for `Enum.Parse` with customized names? I don't think the current implementation accounts for composite values in `[Flags]` enums. The built-in enum converter in STJ does do this [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Converters/Value/EnumConverter.cs) in case you're looking for a reference implementation.

AArnott: I don't see how composite values are necessarily relevant. The names are unique and each have a value assigned, even if the value is not unique. Yes, during deserialization if you encounter a name, you can translate that to a number using this dictionary. Arbitrary combinations of flags (if the enum is flags) is the responsibility of the serializer -- not this map, IMO.

AArnott: I don't see how composite values are necessarily relevant. The names are unique and each have a value assigned, even if the value is not unique. Yes, during deserialization if you encounter a name, you can translate that to a number using this dictionary. Arbitrary combinations of flags (if the enum is flags) is the responsibility of the serializer -- not this map, IMO.

eiriktsarpalis: > Arbitrary combinations of flags (if the enum is flags) is the responsibility of the serializer -- not this map, IMO.

You're possibly right, although I would argue that handling combinations this way is common enough and complex enough to warrant a centralized solution. (Although that could be expressed as a helper building on top of this primitive).

eiriktsarpalis: > IReadOnlyDictionary<string, TUnderlying>

Could this be expressed as a delegate type instead? It would allow for `ReadOnlySpan<char>` based lookups as well as simplify source gen for enums containing a small number of fields.

AArnott: Given it _can_ be added on top of this, I'd prefer to leave it out, at least for now. 
If serialization is enough of a focus for PolyType to merit specialized code for it, and if it can add value by providing helper methods for recognizing how a value is composed of several named values _without_ making any assumptions about how those multiple names will be strung together (e.g. commas, spaces, pipes, etc) then adding it (later?) seems fine to me. 
For my part though, I'm rather content to emit a single name, or an ordinal value in msgpack

AArnott: Regarding delegates, I want to have access to the whole dictionary so I can enumerate possibilities and create my own lookup table based on UTF8 encoded characters.

eiriktsarpalis: Makes sense.

src/PolyType/Abstractions/IEnumTypeShape.cs
```diff
@@ -20,9 +20,20 @@ public interface IEnumTypeShape : ITypeShape
 [InternalImplementationsOnly]
 public interface IEnumTypeShape<TEnum, TUnderlying> : ITypeShape<TEnum>, IEnumTypeShape
     where TEnum : struct, Enum
+    where TUnderlying : unmanaged
```
eiriktsarpalis, line 23: Seems like a reasonable constraint to add, although I'm curious what motivated its inclusion.

src/PolyType/Abstractions/IEnumTypeShape.cs
eiriktsarpalis: > Given it can be added on top of this, I'd prefer to leave it out, at least for now.

I think we'd want to add a few parsing and unparsing helpers under `PolyType.Utils` since pretty much any application has the onus of supporting enums with customized names, including `Flags` enums.

AArnott: Okay I'll give it a shot.

src/PolyType/Abstractions/IEnumTypeShape.cs
AArnott: I figured you might be. :)
It was required as part of the acceptance test. The CborSerializer needed a reverse-lookup table, which meant defining a `Dictionary<TUnderlying, string>` type. But `Dictionary` wants `notnull` applied to its keys. Rather than constrain `TUnderlying` merely to `notnull`, I first used `struct`, then realized `unmanaged` is the most constraining I could do, and yet guaranteed to not be overly-constraining.

src/PolyType/Abstractions/IEnumTypeShape.cs
AArnott: Are you thinking something like this?
https://github.com/AArnott/PolyType/pull/2/files

Note I avoid actually formatting strings, because how JSON formats a multi-flag enum isn't necessarily how another serializer would do it. I also don't parse strings because that's a very simple matter outside of the deserializer-specific splitting of the names.
So these utility methods will tell you whether an enum value matches a well-defined member (or members), and it will break up that value into its members and even returning the remaining bits that didn't fit into any member.

eiriktsarpalis: Seems like a good foundation. Will you be able to dogfood these routines in NB.MessagePack?

AArnott: No, I don't think so. All of these are based on strings and UTF-16 characters. MessagePack is based on UTF-8. The `IsDefinedValueOrCombinationOfValues` method is independent of character encoding, but it only returns a `bool` which is of no use to me. Also, this is all using a case-sensitive dictionary, but Nerdbank.MessagePack uses case-insensitive dictionaries _unless_ the enum actually specifies value names that differ only in capitalization.

(This is why I'm not particularly enthusiastic of the idea of these utility methods. They are special purpose for serializers, yet they may not even be suitable for them because of the specific requirements that serializers have.)

PR #187, opened 2025-06-18
Revise collection construction options to be at activation time

### LINKED ISSUE #139: Consider allowing an optional `capacity` parameter in default collection constructors
Many mutable collection constructors accept a `capacity` parameter, so it might make sense to surface an optional parameter so that users have more control over the initial size. Related to [this conversation](https://github.com/eiriktsarpalis/PolyType/issues/33#issuecomment-2756733798).

cc @drewnoakes @AArnott

AArnott:
This changes how collections are constructed. Instead of exchanging a comparer for a collection factory, you get a collection factory that takes an options struct that may contain a comparer and/or a capacity integer.

* [x] Get sign-off on public API changes.
* [x] Update sourcegen shape provider.
* [x] Update reflection shape provider.
* [x] Update reflection-emit shape provider.
* [x] Rename all "DefaultConstructor" APIs to "MutableConstructor" in the context of collections.
* [x] ~~Add support for capacity property.~~ (this will come in a subsequent change)
* [x] Avoid runtime null check for collection types we expect to support null equality comparer arguments.

Prerequisite to #139


AArnott:
@eiriktsarpalis I'd very much like to get your sign-off on the API before I do anything more.

src/PolyType/Abstractions/CollectionConstructionOptions.cs
```diff
@@ -4,7 +4,7 @@
 /// Options for constructing collections.
 /// </summary>
 /// <typeparam name="TKey">The type of keys within the collection. When collections aren't keyed, this is the element type directly.</typeparam>
-public class CollectionConstructionOptions<TKey>
+public record struct CollectionConstructionOptions<TKey>
```
eiriktsarpalis: In MCP we have started removing records because they were contributing substantially to the static footprint of the application. I expect this would be even more pronounced for the case of generic records.

src/PolyType/Abstractions/CollectionConstructionOptions.cs
```diff
@@ -15,4 +15,9 @@ public class CollectionConstructionOptions<TKey>
     /// Gets an optional comparer for the keys or elements in the collection.
     /// </summary>
     public IComparer<TKey>? Comparer { get; init; }
-}
\ No newline at end of file
+
+    /// <summary>
+    /// Gets the initial capacity of the collection.
```
eiriktsarpalis: Maybe add a remark that this is only honored for the case of default constructors.

src/PolyType/Abstractions/Delegates.cs
```diff
@@ -35,8 +35,13 @@ namespace PolyType.Abstractions;
 /// <typeparam name="T">The element type of the span parameters.</typeparam>
 /// <typeparam name="TDeclaringType">The type of the value produced by the constructor.</typeparam>
 /// <param name="values">The span of values used to create the instance.</param>
+/// <param name="options">An optional set of parameters used to create the collection.</param>
 /// <returns>A newly constructed instance using the specified values.</returns>
-public delegate TDeclaringType SpanConstructor<T, TDeclaringType>(ReadOnlySpan<T> values);
+public delegate TDeclaringType SpanConstructor<T, TDeclaringType>(ReadOnlySpan<T> values, in CollectionConstructionOptions<T>? options = null);
+
+public delegate TDeclaringType DefaultCollectionConstructor<T, TDeclaringType>(in CollectionConstructionOptions<T>? options = null);
```
eiriktsarpalis: Consider calling it `MutableCollectionConstructor` to echo `CollectionConstructionStrategy` nomenclature.

src/PolyType/Abstractions/CollectionConstructionOptions.cs
AArnott: True. I did the same in Nerdbank.MessagePack. In this case I thought it would be common for converters to want to use `options with { Capacity: 5 }` in order to keep the comparer that was previously configured. But that's probably not worth the cost for a generic type, which I hadn't considered. I'll change this back to a regular `struct`.

src/PolyType/Abstractions/CollectionConstructionOptions.cs
AArnott: Interesting I hadn't thought of it as only applicable to default constructors. But looking at the APIs in .net now, I realize that maybe you're right. But that feels more like a consequence of the collection API, rather than a limitation that we might put directly into polytype. After all, a collection constructed that takes an IEnumerable as well as a capacity, could theoretically exist. 
I will certainly change the documentation to indicate it'll only be used when the collection offers a constructor that takes capacity.

AArnott:
All I've got left is the actual Capacity feature work. Everything up to this point was just moving the `CollectionConstructionOptions<TKey>` from 'visitor time' to 'activation time'.

So now that I'm there... I'm asking myself how to recognize the `capacity` parameter in the constructor. That it is an `int` type probably isn't enough, because when the element type is `int`, the constructor may take 1 element of that type (e.g. `ImmutableArray.Create<int>(int)`). 
So perhaps we should match on a parameter being `int capacity` (name and type).

src/PolyType/Abstractions/CollectionConstructionOptions.cs
eiriktsarpalis: Yeah.. Would be nice if C# allowed `with` expressions for non-record types that have copy constructors.

src/PolyType/Abstractions/CollectionConstructionOptions.cs
AArnott: I added a general remarks paragraph about how collections may ignore any/all properties. Capacity isn't unique in that way, and theoretically we may honor capacity for IEnumerable collections any time the API allows it for efficiency gains.

src/PolyType/Abstractions/CollectionConstructionOptions.cs
AArnott: I've dropped the keyword.

src/PolyType.Examples/CborSerializer/CborSerializer.Builder.cs
```diff
@@ -92,7 +92,7 @@ public CborConverter<T> GetOrAddConverter<T>(ITypeShape<T> typeShape) =>
                     new CborMutableEnumerableConverter<TEnumerable, TElement>(
                         elementConverter,
                         getEnumerable,
-                        enumerableShape.GetDefaultConstructor(),
+                        enumerableShape.GetMutableConstructor(),
```
eiriktsarpalis: How about calling it?
```suggestion
                        enumerableShape.GetMutableCollectionConstructor(),
```
Slightly more verbose, but has better self-documentation properties.

src/PolyType/Abstractions/Delegates.cs
```diff
@@ -32,11 +32,33 @@ namespace PolyType.Abstractions;
...
 /// </summary>
-/// <typeparam name="T">The element type of the span parameters.</typeparam>
+/// <typeparam name="TKey">The type to be compared within the <typeparamref name="TElement"/>.</typeparam>
+/// <typeparam name="TElement">The element type of the span parameters.</typeparam>
 /// <typeparam name="TDeclaringType">The type of the value produced by the constructor.</typeparam>
 /// <param name="values">The span of values used to create the instance.</param>
+/// <param name="options">An optional set of parameters used to create the collection.</param>
 /// <returns>A newly constructed instance using the specified values.</returns>
-public delegate TDeclaringType SpanConstructor<T, TDeclaringType>(ReadOnlySpan<T> values);
+public delegate TDeclaringType SpanConstructor<TKey, TElement, TDeclaringType>(ReadOnlySpan<TElement> values, in CollectionConstructionOptions<TKey>? options = null);
```
eiriktsarpalis: Should this be renamed to be consistent with the newly introduced delegate?
```suggestion
public delegate TDeclaringType SpanCollectionConstructor<TKey, TElement, TDeclaringType>(ReadOnlySpan<TElement> values, in CollectionConstructionOptions<TKey>? options = null);
```

src/PolyType/Abstractions/Delegates.cs
```diff
@@ -32,11 +32,33 @@ namespace PolyType.Abstractions;
...
+/// <summary>
+/// Delegate representing a constructor for a collection from an enumeration of elements.
+/// </summary>
+/// <typeparam name="TKey">The type of keys or elements in the collection.</typeparam>
+/// <typeparam name="TElement">The element type of the enumerable.</typeparam>
+/// <typeparam name="TDeclaringType">The type of the collection to be constructed.</typeparam>
+/// <param name="elements">The enumerable of elements used to create the collection.</param>
+/// <param name="options">An optional set of parameters used to create the collection.</param>
+/// <returns>A newly constructed collection instance containing the specified elements.</returns>
+public delegate TDeclaringType EnumerableCollectionConstructor<TKey, TElement, TDeclaringType>(IEnumerable<TElement> elements, in CollectionConstructionOptions<TKey>? options = null);
```
eiriktsarpalis: Unrelated to your changes as such, but I would be curious to hear your thoughts on whether it makes sense to consolidate `EnumerableCollectionConstructor` and `SpanCollectionConstructor` into a single delegate accepting `ReadOnlyMemory<T>` or `ArraySegment<T>`?

Asking because in the example implementations I'm invariably using pooled buffers for both kinds. There is an argument to be made about span-based constructors offering static guarantees on buffer ownership, but I suspect that could be hidden behind the delegate implementation itself: for constructors that accept span or collections that are known to copy the IEnumerable we pass the buffer as-is; for unknown IEnumerable constructors we create defensive copies of the buffer.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -34,9 +34,10 @@ public interface IDictionaryTypeShape : ITypeShape
     CollectionConstructionStrategy ConstructionStrategy { get; }
 
     /// <summary>
-    /// Gets the kind of custom comparer (if any) that this collection may be initialized with.
+    /// Gets an enumeration that describes the properties in <see cref="CollectionConstructionOptions{TKey}"/>
+    /// that this collection supports.
     /// </summary>
-    CollectionComparerOptions ComparerOptions { get; }
+    SupportedCollectionConstructionOptions SupportedConstructionOptions { get; }
```
eiriktsarpalis: I think we should just be opinionated and offer a single option per type, with a strong preference for equality comparer over comparer.

src/PolyType/Abstractions/SupportedCollectionConstructionOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// The collection accepts a custom <see cref="IComparer{T}"/> for sorting and/or equality checks.
+    /// </summary>
+    /// <seealso cref="CollectionConstructionOptions{TKey}.Comparer"/>
+    Comparer,
+
+    /// <summary>
+    /// The collection can be constructed with a custom initial capacity.
+    /// </summary>
+    /// <seealso cref="CollectionConstructionOptions{TKey}.Capacity"/>
+    Capacity,
```
eiriktsarpalis: I think I prefer the previous design of only advertising supported comparers in this enum. 

I don't think the shape needs to report any support for capacity, constructors typically only treat this as a hint that may or may not be ignored. If we absolutely must report this capability, I think I would prefer having it as a separate boolean.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
AArnott: I'm thinking it's totally up to the app to decide whether they want a sorted dictionary or a hashed one for a given `ReadOnlyDictionary`. If at the PolyType level we just force it to be hashed, that just makes PolyType not appropriate for that application, which seems a shame. 
The only part I struggled with was that we _do_ need to define behavior when both types are allowed and both comparers are given (a situation that I doubt will happen in real life), so I stuck my neck out and said it would be hash-based if 0 or 2 comparers are provided.

src/PolyType/Abstractions/SupportedCollectionConstructionOptions.cs
AArnott: Ok, that's fair.

AArnott: Do you want to double-down on this and rename the flags enum back to a Comparer-specific name? While I can't think of what else we'd add to the options struct, theoretically if something comes up and we add it, we might want to add a flag to this supported options enum.

src/PolyType/Abstractions/Delegates.cs
AArnott: Span construction doesn't work with the non-Emit Reflection provider, right? If we removed Enumerable in favor of just Span, wouldn't that be a substantial functionality loss for the non-Emit Reflection crowd?

src/PolyType/Abstractions/IDictionaryTypeShape.cs
eiriktsarpalis: > If at the PolyType level we just force it to be hashed, that just makes PolyType not appropriate for that application, which seems a shame.

Why? If the application needs to construct a ReadOnlyDictionary and we provide one with sound key equality semantics, whether it uses hashing or sorting under the hood is implementation detail, no? Assuming there existed a strong need to change the underlying implementation, a converter can always be special cased at the visitor level by doing `if (typeof(TDictionary) == typeof(ReadOnlyDictionary<TKey, TValue>) ...`.

src/PolyType/Abstractions/Delegates.cs
AArnott: Oh, you're not saying to leave Span as-is, you're talking about consolidating Span and Enumerable into a Memory based one. 
This doesn't come up for MessagePack which always has prefixed lengths, but for collections of unknown size in other formats (cbor, json), not knowing the size up front would naturally mean using `List<T>`. But I can't get a ReadOnlyMemory from a list. I can only get a `Span<T>` for it.

eiriktsarpalis: I'm suggesting we have only one delegate accepting `ReadOnlyMemory<T>` or `ArraySegment<T>`. The non-emit provider should be able to handle those types.

AArnott: That said, I like the idea of reducing 3 collection strategies to 2. If we could collapse span and enumerable into just span without losing no-emit Ref support -- or if we had another general purpose 'growing' collection that would expose a ReadOnlyMemory<T> of itself, that sounds like a good idea.

eiriktsarpalis: I like that idea. There are a few tricks we could use here, including this one from Linq:

https://github.com/dotnet/runtime/blob/f6b13d816ea4ca5ba6aaaa5f3f08f20b6fadd3d8/src/libraries/System.Linq/src/System/Linq/SegmentedArrayBuilder.cs#L15

src/PolyType/Abstractions/IDictionaryTypeShape.cs
AArnott: > whether it uses hashing or sorting under the hood is implementation detail, no?

Never in my experience has it been an implementation detail. I almost always use hashing collections. When I use sorted collections, it's because _I want them sorted_. Typically because I'll be enumerating them and need them to enumerated in a sorted order. If I create a sorted collection and wrap it in a read-only collection wrapper (that itself doesn't know about hashing or sorting), I'd like a deserializer to be able to recreate that.

Now, for this to be a complete story, the deserializer would need to expose a way for the app author to conveniently specify the comparer to use on the property that stores that collection. That seems easy enough:

```cs
[CustomComparer(typeof(MySortingComparer))]
public ReadOnlyDictionary<int, bool> SomeDict;
```

Nerdbank.MessagePack could declare and document that attribute as the way to specify any custom comparer (hash or sorting) and the visitor it defines would find that attribute and arrange for that comparer to be provided via the collection options struct when activating the collection.

Such an attribute would be useful for the non-ambiguous collection types as well, such as `SortedSet<T>` or `HashSet<T>`.

src/PolyType/Abstractions/SupportedCollectionConstructionOptions.cs
eiriktsarpalis: Yeah, I think so.

AArnott: Done.

src/PolyType/Abstractions/IDictionaryTypeShape.cs
eiriktsarpalis: > Never in my experience has it been an implementation detail. I almost always use hashing collections. When I use sorted collections, it's because _I want them sorted_. Typically because I'll be enumerating them and need them to enumerated in a sorted order.

Presumably though this would involve using a type like `SortedDictionary` where sortedness is part of the contract. `ReadOnlyDictionary` doesn't provide any such guarantees.

AArnott: >`ReadOnlyDictionary` doesn't provide any such guarantees.

That's entirely dependent on how it is created. It provides the same guarantees as whatever it's wrapping.

AArnott: Although I suppose if I'm wrapping a sorted collection under a ReadOnly* wrapper, it's to hand off to someone that I don't want messing with my collection. And whether that receiver can depend on the collection retaining the same sorting rules is less clear.

But then, why would someone _serialize_ such a readonly collection? Presumably they were either given it, or maybe they initialized it in their ctor and never want it changed. But _someone_ at some point decided whether it should be sorted, and if so, I imagine they did so intentionally.

eiriktsarpalis: That sounds like a valid concern if you're trying to expose or serialize sorted data that you define and own, but I don't think it transfers to deserialization. When deserializing, you are asking a library to fill a ReadOnlyDictionary shaped hole from raw data, typically without recourse to how that dictionary is created. 

I think this is similar to expecting that deserialized `IEnumerable<T>` values must always be implemented as `List<T>`. If the final result ends up not matching expectations, I would pin this to being a modelling error (using `ReadOnlyDictionary` instead of `SortedDictionary`).

AArnott: If I serialize an `IEnumerable<T>`, I don't expect any particular concrete implementation when I deserialize. But I _do_ expect the order of elements to remain the same.
Sets and dictionaries in general are not as expressly ordered either. But IEnumerable itself implies an order, whether or not its backing store is actually ordered. ReadOnlyDictionary is like that: it retains the order of its underlying data store, if that is ordered.

eiriktsarpalis: > If I serialize an `IEnumerable<T>`, I don't expect any particular concrete implementation when I deserialize. But I _do_ expect the order of elements to remain the same.

That's not a given for arbitrary `IEnumerable<T>` types, unless they also implement `IList<T>`. `HashSet<T>`, `IDictionary<K,V>`, `BlockingCollection<T>` all implement `IEnumerable<T>` but explicitly provide no guarantees on ordering. 

> ReadOnlyDictionary is like that: it retains the order of its underlying data store, if that is ordered.

If it helps convincing you, enumerating a `Dictionary<K,V>` preserves insertion order provided that _no removals are made_. It follows that deserializing a `ReadOnlyDictionary` by wrapping a `Dictionary<K,V>` will preserve ordering as specified in the encoding medium.

AArnott: >That's not a given for arbitrary `IEnumerable<T>` types,

True, but the contract itself provides an ordered sequence (even if it isn't _sorted_, and even if the data store itself isn't sorted). I think we generally expect that enumerating that twice in a row will produce elements in the same order. My point though is that these abstractions don't themselves guarantee a particular order, but they don't mess with it either, and they all have underlying data structures that dictate what order ends up coming out. ReadOnlyDictionary and IEnumerable both behave this way.

>If it helps convincing you, enumerating a `Dictionary<K,V>` preserves insertion order provided that no removals are made.

That's interesting. I was not aware of that guarantee (and my memory suggests it isn't true, but maybe I'm wrong). If that's true, then yes that helps a lot, since deserializing into a `ReadOnlyDictionary` means that no removals will ever happen, and therefore the order _will_ be preserved.

eiriktsarpalis: > I think we generally expect that enumerating that twice in a row will produce elements in the same order.

Would need to double check, but if memory serves BlockingCollection<T> doesn't guarantee this.

> That's interesting. I was not aware of that guarantee (and my memory suggests it isn't true, but maybe I'm wrong).

It's something we had to contend with recently when evaluating potential use of a vectorized hashing algorithm. Because it ended up violating this invariant, it also broke a bunch of our own code that was implicitly relying on it. We decided that keeping that was more important than improving lookup performance. It means that `Dictionary<K,V>` is more useful when modelling DTOs.

AArnott: >but if memory serves BlockingCollection doesn't guarantee this.

Well, that's a multi-threaded work queue, so I expect if it offers a thread-safe, non-blocking enumeration, then of course its contents may change (more even than just reordering) between enumerations.

> We decided that keeping that was more important than improving lookup performance

Good to know. I also whipped up a quick test and what you describes seems to be true. 
In that case, I'll remove the Flags and go back to just preferring hashed based for this as you suggested.

src/PolyType/Abstractions/Delegates.cs
AArnott: Resolving conversation in favor of the issue you filed to track separately.

AArnott:
All feedback addressed.

AArnott:
PolyType Tests are unstable as of the current version. Particularly when running in Release config, they have a tendency to fail. We must have a race condition somewhere. I'm tracking it down. Let's not merge this yet.

AArnott:
The unstable tests turn out to be because of an instability in STJ -- not PolyType. The serialized JSON strings are compared, and PolyType is consistently in the expected order, but STJ sometimes mixed the properties into an unexpected order.

So it's not a problem with this PR. I don't know how you want to fix this. It only repros (so far) on net472. I don't know if that's an indication of where in STJ the unstable behavior comes from. Maybe we should lock in the "last known good" JSON strings and use those as the baseline instead of asking STJ to serialize each time.

src/PolyType/Abstractions/CollectionConstructionOptions.cs
```diff
@@ -4,7 +4,15 @@
...
-public class CollectionConstructionOptions<TKey>
+/// <remarks>
+/// <para>
+/// Construction of any particular collection type may ignore any or all of these properties.
+/// To predict whether a collection will use a particular comparer property, check the
+/// <see cref="CollectionComparerOptions"/> enum, as defined by either
+/// <see cref="IEnumerableTypeShape.SupportedComparers"/> or <see cref="IDictionaryTypeShape.SupportedComparers"/>.
+/// </para>
+/// </remarks>
+public struct CollectionConstructionOptions<TKey>
```
eiriktsarpalis:
```suggestion
public readonly struct CollectionConstructionOptions<TKey>
```

src/PolyType/Abstractions/Delegates.cs
```diff
@@ -32,11 +32,33 @@ namespace PolyType.Abstractions;
...
 /// </summary>
-/// <typeparam name="T">The element type of the span parameters.</typeparam>
+/// <typeparam name="TKey">The type to be compared within the <typeparamref name="TElement"/>.</typeparam>
+/// <typeparam name="TElement">The element type of the span parameters.</typeparam>
 /// <typeparam name="TDeclaringType">The type of the value produced by the constructor.</typeparam>
 /// <param name="values">The span of values used to create the instance.</param>
+/// <param name="options">An optional set of parameters used to create the collection.</param>
 /// <returns>A newly constructed instance using the specified values.</returns>
-public delegate TDeclaringType SpanConstructor<T, TDeclaringType>(ReadOnlySpan<T> values);
+public delegate TDeclaringType SpanCollectionConstructor<TKey, TElement, TDeclaringType>(ReadOnlySpan<TElement> values, in CollectionConstructionOptions<TKey>? options = null);
```
eiriktsarpalis: Can we avoid using nullable here?
```suggestion
public delegate TDeclaringType SpanCollectionConstructor<TKey, TElement, TDeclaringType>(ReadOnlySpan<TElement> values, in CollectionConstructionOptions<TKey> options = default);
```

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -34,9 +34,9 @@ public interface IDictionaryTypeShape : ITypeShape
     CollectionConstructionStrategy ConstructionStrategy { get; }
 
     /// <summary>
-    /// Gets the kind of custom comparer (if any) that this collection may be initialized with.
+    /// Gets the kind of custom comparer(s) (if any) that this collection may be initialized with.
```
eiriktsarpalis: Do we need to revert this change?

src/PolyType/Abstractions/IEnumerableTypeShape.cs
```diff
@@ -25,9 +25,9 @@ public interface IEnumerableTypeShape : ITypeShape
     CollectionConstructionStrategy ConstructionStrategy { get; }
 
     /// <summary>
-    /// Gets the kind of custom comparer (if any) that this collection may be initialized with.
+    /// Gets the kind of custom comparer(s) (if any) that this collection may be initialized with.
```
eiriktsarpalis: Likewise

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -81,10 +81,9 @@ public interface IDictionaryTypeShape<TDictionary, TKey, TValue> : ITypeShape<TD
     /// <summary>
     /// Creates a delegate wrapping a parameterless constructor of a mutable collection.
     /// </summary>
-    /// <param name="collectionConstructionOptions">Options that control how the collection is constructed. Use <see cref="IDictionaryTypeShape.ComparerOptions" /> to predict which properties may be worth initializing.</param>
     /// <exception cref="InvalidOperationException">The collection is not <see cref="CollectionConstructionStrategy.Mutable"/>.</exception>
-    /// <returns>A delegate wrapping a default constructor.</returns>
-    Func<TDictionary> GetDefaultConstructor(CollectionConstructionOptions<TKey>? collectionConstructionOptions = null);
+    /// <returns>A delegate wrapping a constructor that takes no elements.</returns>
```
eiriktsarpalis:
```suggestion
    /// <returns>A delegate creating an empty mutable collection.</returns>
```

src/PolyType/Abstractions/IDictionaryTypeShape.cs
```diff
@@ -81,10 +81,9 @@ public interface IDictionaryTypeShape<TDictionary, TKey, TValue> : ITypeShape<TD
     /// <summary>
     /// Creates a delegate wrapping a parameterless constructor of a mutable collection.
```
eiriktsarpalis:
```suggestion
    /// Creates a delegate creating an empty mutable collection.
```

src/PolyType/Abstractions/Delegates.cs
AArnott: Oh! I hadn't thought of that. That will make some other things slightly simpler too.

eiriktsarpalis:
Thanks

PR #193, opened 2025-06-25
Add support for unity

### LINKED ISSUE #191: Support for Unity
Unity doesn't have C# 12 support yet.

If support for generic attributes is the only reason C# 12 is required, what about adding a `GenerateShapeAttribute(Type)` alternative to `GenerateShapeAttribute<T>`? 

I feel like we explored this before, but I don't remember what came of it.

AArnott:
This adds support for unity by:

1. Restricting generated code to C# 9.
2. Retargets the source generator for the Roslyn 4.3 compiler.
3. This adds a `GenerateShapeAttribute(Type)` constructor, so that witness types can be generated without requiring the C# 12 support for generic attributes.

This will remove the very hacky step Unity users currently have to use to replace their compiler version to one that supports C# 12, as Nerdbank.MessagePack currently documents: 

> Follow the [instructions on UnityRoslynUpdater](https://github.com/DaZombieKiller/UnityRoslynUpdater?tab=readme-ov-file#using-c-12) to enable C# 12+ language versions for your project.

Closes #191

### DIFF
--- a/tests/PolyType.Tests/CollectionsWithCapacityTests.cs
+++ b/tests/PolyType.Tests/CollectionsWithCapacityTests.cs
@@ -42,10 +42,10 @@ public abstract partial class CollectionsWithCapacityTests(ProviderUnderTest pro
 #endif
     }
 
-    [GenerateShape<List<int>>]
-    [GenerateShape<List<string>>]
-    [GenerateShape<HashSet<int>>]
-    [GenerateShape<Dictionary<int, bool>>]
+    [GenerateShapeFor<List<int>>]
+    [GenerateShapeFor<List<string>>]
+    [GenerateShapeFor<HashSet<int>>]
+    [GenerateShapeFor<Dictionary<int, bool>>]
     partial class Witness;
 
     ////public sealed class Reflection() : CollectionsWithCapacityTests(ReflectionProviderUnderTest.NoEmit);


src/PolyType.SourceGenerator/Analyzers/TypeShapeExtensionAnalyzer.cs
```diff
@@ -12,7 +15,11 @@ public class TypeShapeExtensionAnalyzer : DiagnosticAnalyzer
 
     public override void Initialize(AnalysisContext context)
     {
-        context.EnableConcurrentExecution();
+        if (!Debugger.IsAttached)
```
eiriktsarpalis, line 18: Why is this check necessary now?

src/PolyType.SourceGenerator/Analyzers/TypeShapeExtensionAnalyzer.cs
```diff
@@ -25,47 +32,75 @@ public override void Initialize(AnalysisContext context)
                 }
 
                 Dictionary<INamedTypeSymbol, (INamedTypeSymbol Marshaller, Location AttributeLocation)> extensionMarshaler = new(SymbolEqualityComparer.Default);
-                context.RegisterOperationAction(
+                context.RegisterSyntaxNodeAction(
```
eiriktsarpalis, line 35: Will we need to revert once we move to a newer version of Roslyn again?

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.Constructors.cs
```diff
@@ -51,8 +53,8 @@ static string FormatAttributeProviderFunc(ObjectShapeModel type, ConstructorShap
             }
 
             string parameterTypes = constructor.Parameters.Length == 0
-                ? "[]"
-                : $$"""[{{string.Join(", ", constructor.Parameters.Select(p => $"typeof({p.ParameterType.FullyQualifiedName})"))}}]""";
+                ? "global::System.Type.EmptyTypes"
```
eiriktsarpalis, line 56: Why not just `Array.Empty<Type>()`?

src/PolyType.SourceGenerator/Analyzers/TypeShapeExtensionAnalyzer.cs
AArnott: It isn't. But I find it immensely helpful. It doesn't impact shipping behavior but it makes debugging these analyzers much easier. I can remove it if you don't want it.

src/PolyType/GenerateShapeAttribute.cs
```diff
@@ -3,19 +3,54 @@
...
 /// Projects targeting older versions of .NET need to access the generated
 /// <see cref="ITypeShapeProvider"/> instance through the static property
-/// added to classes annotated with the <see cref="GenerateShapeAttribute{T}"/>.
+/// added to classes annotated with the <see cref="GenerateShapeAttribute{T}"/>
+/// or <see cref="GenerateShapeAttribute(Type)"/>.
+/// </para>
 /// </remarks>
-[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Interface, AllowMultiple = false, Inherited = false)]
-public sealed class GenerateShapeAttribute : Attribute;
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Interface, AllowMultiple = true, Inherited = false)]
```
eiriktsarpalis: Why `AllowMultiple = true`?

src/PolyType/GenerateShapeAttribute.cs
```diff
@@ -3,19 +3,54 @@
...
+[Conditional("NEVER")] // only the source generator uses this.
```
eiriktsarpalis, line 19: What does this do?

src/PolyType.SourceGenerator/Analyzers/TypeShapeExtensionAnalyzer.cs
AArnott: No. This will work either way. The operation syntax was perhaps simpler. It might even be more performant, though I've done what I can to order the checks in this delegate so that we only access the semantic model after all the cheap checks are done, which should keep the number of times the semantic model is accessed to just the number of assembly-level attributes that exist, which are typically very few.

src/PolyType/GenerateShapeAttribute.cs
eiriktsarpalis: Oh I see, this is repurposing the same attribute we use for IShapeable<T> generation? Hmm...

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.Constructors.cs
AArnott: Because this is generated code. So to be bullet-proof it would have to be 

`global::System.Array.Empty<global::System.Type>()`
which is quite a bit longer than
`global::System.Type.EmptyTypes`

And anyway, this is exactly what `Type.EmptyTypes` was designed for. :)

src/PolyType/GenerateShapeAttribute.cs
eiriktsarpalis: Can we somehow keep those responsibilities distinct? E.g. by introducing a new attribute with a separate name and/or renaming some of the existing ones?

tests/PolyType.SourceGenerator.UnitTests/Analyzers/TypeShapeExtensionAnalyzerTests.cs
```diff
@@ -24,7 +24,12 @@ public class PointMarshaller : IMarshaller<Point, PointMarshaller.PointSurrogate
 
                 public PointSurrogate ToSurrogate(Point value) => throw new System.NotImplementedException();
 
-                public record struct PointSurrogate(int X, int Y);
+                public struct PointSurrogate
```
eiriktsarpalis, line 27: Is this changed because C# 9 does't support records? We should still be able to up the language version in testing right?

src/PolyType/GenerateShapeAttribute.cs
AArnott: Because this has to be allowed:

```cs
[GenerateShape(typeof(Foo))]
[GenerateShape(typeof(Bar))]
partial class Witness;
```

Before, with no constructor parameters, there was never a need for the attribute to appear multiple times.

The only way to avoid this is to give the witness-targeted attribute a new name. e.g. `GenerateWitnessShapeForAttribute<T>` and `GenerateWitnessShapeForAttribute(Type)`. I elected to not go for that. But if you'd prefer that, we can do that.

src/PolyType/GenerateShapeAttribute.cs
AArnott: Same thing as for `GenerateShapeAttribute<T>`, which already has it: it removes it from the compiled output. It is never used at runtime, so it makes the output slightly smaller.

This was particularly important for `GenerateShapeAttribute<T>` so that .NET Framework doesn't throw when reflecting over generic attributes (which it doesn't support). The non-generic `GenerateShapeAttribute` doesn't _need_ this, but it helps to emphasize that it is only a source generation tool.

tests/PolyType.SourceGenerator.UnitTests/Analyzers/TypeShapeExtensionAnalyzerTests.cs
AArnott: C# 9 only supports `record`. It doesn't support `record class` or `record struct` (even though `record` itself is a class).
So yes, I can (and do!) force the language version higher for some tests whose point is specifically to test newer language syntax. But this test wasn't really about testing `record struct`, so I just reduced the syntax so that it can run under more contexts.

src/PolyType/GenerateShapeAttribute.cs
AArnott: Your follow-up comments appeared only after I submitted my first reply. 
What would you like to rename the generic attribute and this new non-generic witness attribute to?

eiriktsarpalis: Here's one idea:
```
[GenerateShape] // Generates for the current type
public partial class Foo;

[GenerateShapeOf(typeof(Foo))]
[GenerateShapeOf<Foo>]
public partial class Witness;
```
Naming needs a bit of work, but structurally I think this is where we need to head towards.

src/PolyType.SourceGenerator/Analyzers/TypeShapeExtensionAnalyzer.cs
eiriktsarpalis: Makes sense. I thought moving to an earlier version somehow made this necessary.

src/PolyType/GenerateShapeAttribute.cs
AArnott: Renamed to `GenerateShapeFor`.

src/PolyType/GenerateShapeForAttribute.cs
```diff
@@ -0,0 +1,47 @@
...
+/// <para>
+/// For projects targeting .NET 8 or later, this additionally augments the class
+/// with an implementation of IShapeable for <typeparamref name="T"/>.
+/// </para>
+/// </remarks>
+/// <seealso cref="GenerateShapeForAttribute"/>
+/// <seealso cref="GenerateShapeAttribute"/>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, AllowMultiple = true, Inherited = false)]
+[Conditional("NEVER")] // only the source generator uses this, and avoiding generic attributes avoid .NET Framework and Unity issues.
+public sealed class GenerateShapeForAttribute<T> : Attribute;
```
eiriktsarpalis, line 24: Do we want to make this a subclass of the non-generic variant? Should make the reflection-based implementation simpler if anything.

tests/PolyType.Tests/CollectionShapeTests.cs
```diff
@@ -68,7 +68,7 @@ public InternalDictionaryOfIntWithPublicConstructor()
         }
     }
 
-    [GenerateShape<PublicListOfIntWithInternalConstructor>]
+    [GenerateShapeFor<PublicListOfIntWithInternalConstructor>]
```
eiriktsarpalis, line 71: Are there any markdown/docs files that also need updating?

src/PolyType/GenerateShapeForAttribute.cs
AArnott: Reflection-based providers don't use this attribute _at all_. And source generators have to look for each attribute type specifically anyway. Reflection based providers can't see these attributes because they vanish at compile time (due to `Conditional("NEVER")`).
So no, it wouldn't really make anything simpler.

tests/PolyType.Tests/CollectionShapeTests.cs
AArnott: Yes, and done.

PR #198, opened 2025-06-27
Add copy constructor to `CollectionConstructionOptions<TKey>`

AArnott:
Without this, deserializers have to manually copy properties from a 'template' options struct which may have comparers in it into a local variable for the struct in order to set `Capacity`. Since all these properties are declared as `init`, one cannot simply copy the struct into a local and then change Capacity -- it must be done all at construction time, which is tedious. And when we add properties to this struct, all those deserializers will drop that added property during their manual copy.

With the new copy constructor, the following syntax becomes an option:

```cs
CollectionConstructionOptions<TKey> options = new(this.optionsTemplate)
{
   Capacity = count,
};
```

Alternative designs:

1. Use `record struct` so that C# users can say `this.optionsTemplate with { Capacity = count }`. This is the nicest syntax, but adds a lot of generated code to the output, and since this is a generic type, that can bloat the size of the final executable.
2. Change from a `readonly struct` to a mutable one, and the property accessors from `init` to `set`.

### DIFF
--- a/src/PolyType/Abstractions/CollectionConstructionOptions.cs
+++ b/src/PolyType/Abstractions/CollectionConstructionOptions.cs
@@ -15,6 +15,12 @@
 public readonly struct CollectionConstructionOptions<TKey>
 {
     /// <summary>
+    /// Initializes a new instance of the <see cref="CollectionConstructionOptions{TKey}"/> struct.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


eiriktsarpalis (commented):
Looks good, but why would one need a copy constructor if you can simply rely on struct copy semantics?

PR #199, opened 2025-06-29
Suppress CS0612 obsolete warnings

AArnott:
When emitting shapes for types with obsolete properties, CS0612 may be emitted from generated code, with no way to suppress it from the user's project. We should suppress these within the source generator.

### DIFF
--- a/src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.cs
+++ b/src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.cs
@@ -102,6 +102,11 @@ internal sealed partial class SourceFormatter(TypeShapeProviderModel provider)
             """);
 #endif
 
+        writer.WriteLine("""
+            #pragma warning disable CS0612, CS0618 // Use of obsolete APIs is natural when we're emitting delegates for obsolete properties.
+
+            """);
+
         if (typeDeclaration.Namespace is string @namespace)
         {
             writer.WriteLine($"namespace {@namespace}");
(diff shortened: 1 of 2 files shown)


src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.cs
```diff
@@ -102,6 +102,11 @@ private static void StartFormatSourceFile(SourceWriter writer, TypeDeclarationMo
             """);
 #endif
 
+        writer.WriteLine("""
+            #pragma warning disable CS0612 // Use of obsolete APIs is natural when we're emitting delegates for obsolete properties.
```
eiriktsarpalis: This should also include CS0618 which is the separate diagnostic id emitted if an argument is passed to the obsolete attribute. Note also that as part of the "better obsoletions" effort undertaken by the libraries team a few years back we're now allocating [individual diagnostic id's](https://github.com/dotnet/runtime/blob/main/docs/project/list-of-diagnostics.md#obsoletions) for each API that is being obsoleted. This obviously doesn't mix well with source generators, so we have been discussing a potential mechanism for injecting user-defined suppressions into generated code.

AArnott: Added

eiriktsarpalis: You can merge both suppressions in a single line using comma separations.

eiriktsarpalis: https://github.com/dotnet/runtime/blob/da3d9d0f20f5b1b893fa0609e4c37e0eb09dc6ce/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Emitter.cs#L146-L147

AArnott: Oh, right. Good idea.

PR #201, opened 2025-07-16
Rework collection construction resolution implementations.

eiriktsarpalis:
This PR:

1. Reworks the collection constructor resolution algorithms so that a single collection constructor/factory method overload is selected, optimizing for presence of custom comparers, parameter arity and choice of value input type in that order.
2. Implements support for capacity parameters in the reflection-based providers.
3. Adds support for the collection interfaces under System.Collections.Immutable.

cc @AArnott


src/PolyType/ReflectionProvider/MemberAccessors/ReflectionMemberAccessor.cs
```diff
@@ -8,6 +9,7 @@
 namespace PolyType.ReflectionProvider.MemberAccessors;
 
 [RequiresUnreferencedCode(ReflectionTypeShapeProvider.RequiresUnreferencedCodeMessage)]
+[RequiresDynamicCode(ReflectionTypeShapeProvider.RequiresDynamicCodeMessage)]
```
AArnott, line 12: Is this going to cause problems for NativeAOT processes that otherwise could have used the no-emit reflection provider?

eiriktsarpalis: It shouldn't, the reflection provider is already doing a bunch of `MakeGenericType` calls that make supporting such runtimes impossible.

Issue #189, opened 2025-06-23
Consolidate SpanCollectionConstructor with EnumerableCollectionConstructor.


eiriktsarpalis:
That said, I like the idea of reducing 3 collection strategies to 2. If we could collapse span and enumerable into just span without losing no-emit Ref support -- or if we had another general purpose 'growing' collection that would expose a ReadOnlyMemory<T> of itself, that sounds like a good idea.

_Originally posted by @AArnott in https://github.com/eiriktsarpalis/PolyType/pull/187#discussion_r2161862968_

AArnott:
One other thought though: if this growing collection was contractually obligated to be able to produce a `Memory<T>`, that constrains it to store its entire content inside a single array, rather than using other data structures that would avoid copying elements as the structure grows. 
I guess `List<T>` guarantees that it behaves this way already, and everyone uses that, so probably not a big deal.

eiriktsarpalis:
It's a holdover from a similar distinction employed by STJ:

https://github.com/eiriktsarpalis/PolyType/blob/2e59a9db218d8212eeedd63cd9d36847fc4eaef6/src/PolyType/ReflectionProvider/ReflectionTypeShapeProviderOptions.cs#L15-L21

TL;DR we're defaulting to noemit for runtimes not supporting runtime code generation and use frameworks that still don't rely on source generated serialization.

AArnott:
Are there runtimes that do not support runtime code generation that _don't_ also imply trimming?

PR #202, opened 2025-07-17
Deprecate `CollectionConstructionStrategy.Enumerable`

eiriktsarpalis:
This PR deprecates `CollectionConstructionStrategy.Enumerable` and consolidates all parameterized collection construction using span-based delegates, now renamed to `CollectionConstructionStrategy.Parameterized`. This is done for the following reasons:

1. Avoid duplicated code when handling collection shapes.
2. Avoid ambiguity re: ownership of the buffers being passed to the parameterized delegate. Using pooled buffers when building the IEnumerable could be potentially dangerous if a collection attempts to consume the enumerable after construction. By moving everything to span the buffers are unambiguously owned by the caller and can be safely reclaimed once deserialization is complete.
3. With the introduction of dictionary expressions/CollectionBuilderAttribute support for dictionary types, span-based collection initialization is going to become effectively ubiquitous, with very few types needing to convert to intermediate IEnumerables.

This is a breaking change. cc @AArnott


AArnott:
Does this mean #189 is being addressed after all?

eiriktsarpalis:
> Does this mean https://github.com/eiriktsarpalis/PolyType/issues/189 is being addressed after all?

Pretty much, although my original prototype for that was much more elaborate using a special "collection builder" abstraction. It was the work in #201 that convinced that all you really need is spans :-).

PR #209, opened 2025-07-20
Fix concurrency bug in lazy init of object constructor shapes

AArnott:
Fixes https://github.com/AArnott/Nerdbank.MessagePack/issues/485

Another concurrency bug in lazy initialization was causing unstable tests in Nerdbank.MessagePack where the deserializer would incorrectly report that required properties were missing from the deserialized stream when in fact they were present.

Before the fix in addition to tests failing randomly in my CI/PR, I was able to get it to fail using Run Until Failure in VS. The first time it failed on iteration 20. The second time it failed on the first iteration. With the fix, I got through 200 iterations without any failures.

I also improve the debugger display of the properties on these shapes.

### DIFF
--- a/src/PolyType/SourceGenModel/SourceGenObjectTypeShape.cs
+++ b/src/PolyType/SourceGenModel/SourceGenObjectTypeShape.cs
@@ -49,9 +49,9 @@ public sealed class SourceGenObjectTypeShape<TObject> : SourceGenTypeShape<TObje
         {
             if (!_isConstructorResolved)
             {
-                if (CreateConstructorFunc is not null)
+                if (CreateConstructorFunc?.Invoke() is { } constructor)
                 {
-                    CommonHelpers.ExchangeIfNull(ref _constructor, CreateConstructorFunc());
+                    CommonHelpers.ExchangeIfNull(ref _constructor, constructor);
                 }
 
                 Volatile.Write(ref _isConstructorResolved, true);
--- a/src/PolyType/SourceGenModel/SourceGenObjectTypeShape.cs
+++ b/src/PolyType/SourceGenModel/SourceGenObjectTypeShape.cs
@@ -1,4 +1,6 @@
-﻿using PolyType.Abstractions;
+﻿using System.Data;
+using System.Diagnostics;
+using PolyType.Abstractions;
 
 namespace PolyType.SourceGenModel;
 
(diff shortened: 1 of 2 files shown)


src/PolyType/SourceGenModel/SourceGenObjectTypeShape.cs
```diff
@@ -34,23 +35,15 @@ public sealed class SourceGenObjectTypeShape<TObject> : SourceGenTypeShape<TObje
...
-                Volatile.Write(ref _isConstructorResolved, true);
-            }
-
-            return _constructor;
-        }
-    }
-
-    private bool _isConstructorResolved;
+    /// <inheritdoc/>
+    public IConstructorShape? Constructor => _constructor ?? (CreateConstructorFunc is null ? null : CommonHelpers.ExchangeIfNull(ref _constructor, CreateConstructorFunc.Invoke()));
```
eiriktsarpalis: I don't think this will work because `CreateConstructorFunc` can return null, meaning that it will force a recomputation every time that property is being poked. I think we need to keep the original structure but also write to the `_constructor` field using `ExchangeIfNull`. That was my plan when making that refactoring but it seems I forgot to do it.

eiriktsarpalis: > because CreateConstructorFunc can return null

Actually, that isn't the case but I don't think we should rely on the source generator observing that invariant.

AArnott: Really? We do everywhere else:

Image

In fact in the other places, the delegates are allowed to be null as well and yet we invoke the ExchangeIfNull method and 'lock in' a default empty collection in that case. This isn't a collection, but in all other respects it appears to be the same. None of the delegates are annotated as null returning.
And worst case scenario, we have hand-written outside code that improperly returns null from a non-null delegate (already a series of unlikely events), we just get something that calls Interlocked on each call, which considering this shouldn't be in a throughput perf critical area anyway... IMO doesn't justify more complexity.

That said, if you really want me to, I'll bring that back.

src/PolyType/SourceGenModel/SourceGenObjectTypeShape.cs
```diff
@@ -34,16 +36,24 @@ public sealed class SourceGenObjectTypeShape<TObject> : SourceGenTypeShape<TObje
...
     {
         get
         {
             if (!_isConstructorResolved)
             {
-                _constructor = CreateConstructorFunc?.Invoke();
+                if (CreateConstructorFunc is not null)
+                {
+                    CommonHelpers.ExchangeIfNull(ref _constructor, CreateConstructorFunc());
+                }
```
eiriktsarpalis: Minor nitpick:
```suggestion
                if (CreateConstructorFunc?.Invoke() is { } constructor)
                {
                    CommonHelpers.ExchangeIfNull(ref _constructor, constructor);
                }
```

eiriktsarpalis (approved):
Thanks

PR #211, opened 2025-07-22
Allow duplicate element tracking for mutable collections.

### LINKED ISSUE #205: Add support for duplicate dictionary key detection
PT library authors currently have no way of telling if a dictionary is being constructed using duplicate key occurrences, something that can be exploited in the context of [serialization interoperability attacks](https://bishopfox.com/blog/json-interoperability-vulnerabilities?formCode=MG0AV3).  

In the context of mutable dictionaries, the "Add" delegate maps to the type's indexer which usually carries overwrite semantics without providing any feedback on the status of the insertion operation (new write, replacement). In immutable dictionaries, the user is simply passing a span of key/value pairs and delegates the responsibility of duplicate handling to the constructor or factory method: older methods typically throw but newer ones (e.g. .NET 10 CBA factories) have overwrite semantics.

A poor man's solution to the problem is to simply compare the final dictionary count to the number of initial values: if it's smaller, then the payload contains duplicates. While this mitigates the security issue, it is pretty terrible from an error reporting perspective. Another would involve tracking inbound keys in a separate HashSet: while this should work in 90% of cases it is not guaranteed that the equality of the set is going to be a 1:1 correpondence to the equality of the final hydrated dictionary.

For the case of mutable dictionaries specifically, one potential solution would involve mapping the "Add" delegate to `TryAdd` methods (or `ContainsKey`/`Add` where `TryAdd` is not available). We could have either the delegate (or the factory of the delegate) accept an `InsertionBehavior` enum (`Keep`, `Overwrite`) that controls how conflicts are handled. I do not have good solution in mind for parameterized dictionaries, but happy to entertain suggestions.

eiriktsarpalis:
This PR enables optional duplicate element/key tracking for mutable collection types. It does so by replacing the `Setter<,>` delegate used to represent the "Add" functions with bespoke delegate types that return `bool` indicating the success of the operation. The `GetAddMethod()` name has been replaced with `GetAppender()` and `GetInserter()` for enumerable and dictionary types respectively. Dictionary types additionally accept a `DictionaryInsertionMode` enum that controls how duplicate will be handled.

cc @AArnott 

Fix #205.

### DIFF
--- a/src/PolyType/Abstractions/DictionaryInsertionMode.cs
+++ b/src/PolyType/Abstractions/DictionaryInsertionMode.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace PolyType.Abstractions;
+
+/// <summary>
+/// Specifies insertion behavior for a mutable dictionary type.
+/// </summary>
+[Flags]
+public enum DictionaryInsertionMode
+{
+    /// <summary>
+    /// No specific insertion behavior is specified.
+    /// </summary>
+    /// <remarks>
+    /// When passed to <see cref="IDictionaryTypeShape{TDictionary, TKey, TValue}.GetInserter(DictionaryInsertionMode)"/>,
+    /// picks any available insertion method offered by the dictionary.
+    /// </remarks>
+    None = 0,
+
+    /// <summary>
+    /// Overwrites the existing value with the new value when the key already exists.
+    /// </summary>
+    /// <remarks>
+    /// Canonically maps to the indexer of the dictionary.
+    /// </remarks>
+    Overwrite = 1,
+
+    /// <summary>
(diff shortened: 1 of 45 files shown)


src/PolyType/Abstractions/DictionaryInsertionMode.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// </remarks>
+    Discard = 2,
+
+    /// <summary>
+    /// Throws an exception when the key already exists.
+    /// </summary>
+    /// <remarks>
+    /// Canonically maps to the "Add" method of the dictionary.
+    /// </remarks>
+    Throw = 4,
```
AArnott, line 47: nit: using `0x4` instead of `4` (and similar to the above values) screams 'flags enum!' so that folks are less likely to accidentally at `5` here ignorantly.

AArnott (approved):
Looks good.

PR #212, opened 2025-07-26
Introduce an `IArgumentState` interface for constructor shapes.

### LINKED ISSUE #140: Add expose an `AreRequiredParameters` predicate to `IConstructorShape`

eiriktsarpalis:
Introduces an `IArgumentState` interface for constructor shapes that supports tracking of required parameters and detecting duplicate parameter assignments. Introduces a binary but not source breaking change.

Fix #140. cc @AArnott

### DIFF
--- a/src/PolyType.Examples/CborSerializer/CborSerializer.Builder.cs
+++ b/src/PolyType.Examples/CborSerializer/CborSerializer.Builder.cs
@@ -61,7 +61,7 @@ public static partial class CborSerializer
 
             if (constructor.Parameters is [])
             {
-                return new CborObjectConverterWithDefaultCtor<TDeclaringType>(constructor.GetDefaultConstructor(), properties);
+                return new CborObjectConverterWithDefaultCtor<TDeclaringType>(constructor.GetDefaultConstructor(), properties, constructor.DeclaringType.Properties);
             }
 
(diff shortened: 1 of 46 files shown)


AArnott:
Awesome. I will try plugging this into my library and replace what I built and see if I can achieve parody or better behavior.

AArnott:
Migration in progress... noting lack of any additional documentation to talk about how to activate the additional enforcement.

AArnott:
It works for me. And in fact it caused one test failure in my code that so far can only explained as a blatant bug in my own code. So kudos to you!

AArnott (approved):

Issue #206, opened 2025-07-18
Review API naming conventions


eiriktsarpalis:
There exist a few naming inconsistencies in some of the public APIs, e.g. object shapes use the term "default constructor" whereas collection shapes use the term "mutable constructor". We should review the API surface for consistency before making a stable release.

PR #213, opened 2025-07-29
Use consistent terminology for default constructor delegates.

eiriktsarpalis:
Uses consistent terminology in default constructor delegates in both object and collection shapes. I've settled towards keeping `MutableCollectionConstructor` for the delegate type but `GetDefaultConstructor()` for the method that returns the mutable collection constructor. This keeps it consistent with `IConstructorShape` APIs and related helpers.

Contributes to #206.


eiriktsarpalis:
Heads for this breaking change @AArnott.

PR #227, opened 2025-08-08
Use marshalers for converting between union types and union case types.

### LINKED ISSUE #216: Replace the `TUnionCase : Union` constraint on `IUnionCaseShape` with a marshaller
The only instances of union types that PolyType targets today are F# union types and class hierarchies. In both cases the subtype relationship between the union and union case types holds (including, trivially, in the case of struct unions). This however might not hold true with the introduction of union types in C#, where certain lowering strategies do not observe this invariant. For such cases, we need to expose a marshaller for converting between union case types and union instances:

```diff
public partial interface IUnionCaseShape<TUnionCase, TUnion>
-    where TUnionCase : TUnion
{
+    IMarshaller<TUnionCase, TUnion> Marshaller { get; } // 1-1 but not surjective
}
```

cc @AArnott

eiriktsarpalis:
Makes the following changes:

* Updates the `IUnionCaseShape<TUnionCase, TUnion>` type so that conversions between the union case type and the union type happen using a marshaller as opposed to requiring a subtype relationship.
* Reworks the `IMarshaler` abstraction to reflect growing its scope beyond just surrogates. Rename from `Marshaler` to `Marshaller` reflecting American English and .NET conventions.
* Misc cleanups to the source gen model.

This is a breaking change cc @AArnott 

Fix #216.


AArnott:
My US English spell checker wants me to use 'marshaller' rather than 'marshaler'. So it also seems like you went _toward_ British English rather than away from it as you described. Was that also your intent?

PR #229, opened 2025-08-08
Annotate that marshalers should only exchange null for null

AArnott:


src/PolyType/IMarshaler.cs
```diff
@@ -17,12 +19,14 @@ public interface IMarshaler<TSource, TTarget>
     /// </summary>
     /// <param name="value">The input of type <typeparamref name="TSource"/>.</param>
     /// <returns>A value of type <typeparamref name="TTarget"/> corresponding to <paramref name="value"/>.</returns>
+    [return: NotNullIfNotNull(nameof(value))]
```
eiriktsarpalis, line 22: This is effectively equivalent to requiring that both input and output are non-nullable. The current annotation serves two purposes:

1. It enables mapping of null to non-null and vice-versa and
2. From the perspective of consuming library authors it doubles down on the notion that we cannot trust that implementations are honoring nullability annotations.

I think we could do better from the perspective of _authoring_ marshallers. One possibility is exposing an abstract `Marshaller<TSource, TTarget>` class that implements IMarshaller but uses different annotations for its methods and which short circuits null value handling.

AArnott: > This is effectively equivalent to requiring that both input and output are non-nullable

That isn't my intent nor the effect. It allows null, but it means null may only be returned if null was the input.

My intent is to avoid the more complex code I had to write [here](https://github.com/AArnott/Nerdbank.MessagePack/pull/536/files#diff-093aa58c215ee2206853c979b9df747af7f0cd667322b154dce2965dc2fc7cbeR26):

```cs
public int GetHashCode([DisallowNull] TUnion obj) => marshaler.Unmarshal(obj) is TUnionCase c ? inner.GetHashCode(c) : 0;
```

That used to be a simple:

```cs
public int GetHashCode([DisallowNull] TUnion obj) => inner.GetHashCode((TUnionCase)obj);
```

What is the scenario that warrants a marshaller to exchange null for a non-null value? It wasn't allowed before your recent change (where people would just cast), and your change was meant to prepare us for C# discriminated type unions, which also wouldn't need to convert a non-null to a null value AFAIK.

eiriktsarpalis: > > This is effectively equivalent to requiring that both input and output are non-nullable
> 
> That isn't my intent nor the effect. It allows null, but it means null may only be returned if null was the input.

But doesn't it follow necessary when the mapping needs to be bidirectional?

> What is the scenario that warrants a marshaller to exchange null for a non-null value?

In the lowering strategy proposed for [nominal type unions](https://github.com/dotnet/csharplang/blob/main/proposals/nominal-type-unions.md#lowering) it is possible to get back `null` if you decontruct a never-null union type. I don't believe the opposite is possible (unions constructed from non-null union cases should theoretically be non-null, modulo any [optimization hacks](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-compilationrepresentationflags.html#UseNullAsTrueValue) permitted by the compiler(s)). Ultimately though, this annotation is primarily informed by surrogate use cases as opposed to unions. 

Perhaps it makes sense to use a separate interface altogether for unions, so that nullability annotations can be updated independently if such a requirement ever arises.

PR #234, opened 2025-08-18
Suppress CS0436 warning in generated code

AArnott:
In such an event, the user's project _already_ has to deal with such type name collisions and has likely suppressed the warning in their own usages of the types. But they can't do that within generated code, so PolyType should.


src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.cs
```diff
@@ -104,6 +104,7 @@ private static void StartFormatSourceFile(SourceWriter writer, TypeDeclarationMo
 
         writer.WriteLine("""
             #pragma warning disable CS0612, CS0618 // Use of obsolete APIs is natural when we're emitting delegates for obsolete properties.
+            #pragma warning disable CS0436 // Use of local types when imported types by the same name exist.
```
eiriktsarpalis, line 102: I'm not familiar with that warning, any chance you could include a unit test for my edification?

AArnott: Done.

Issue #252, opened 2025-09-06
Source generated type shape provider loads all assemblies for shaped types when only one is requested


AArnott:
In adopting this in Visual Studio, we noticed unexpected assembly loads on startup when migrating to PolyType. Here is what I found as the blame when I wrote an assembly load test:

```
System.Exception : Assembly System.Numerics was loaded unexpectedly by the test with this stack trace: 
   at PolyType.SourceGenerator.ShapeProvider_Nerdbank_MessagePack_Tests.GetShape(Type type)
   at PolyType.Abstractions.TypeShapeProvider.Resolve(ITypeShapeProvider shapeProvider, Type type)
   at PolyType.Utilities.TypeCache.GetOrAdd(Type type)
   at Nerdbank.MessagePack.PolyfillExtensions.GetOrAddOrThrow(MultiProviderTypeCache cache, Type type, ITypeShapeProvider provider) in D:\source\repos\Nerdbank.MessagePack\src\Nerdbank.MessagePack\PolyfillExtensions.cs:line 47
   at Nerdbank.MessagePack.ConverterCache.GetOrAddConverter[T](ITypeShapeProvider provider) in D:\source\repos\Nerdbank.MessagePack\src\Nerdbank.MessagePack\ConverterCache.cs:line 130
   at Nerdbank.MessagePack.MessagePackSerializer.Serialize[T](MessagePackWriter& writer, T& value, ITypeShapeProvider provider, CancellationToken cancellationToken) in D:\source\repos\Nerdbank.MessagePack\src\Nerdbank.MessagePack\MessagePackSerializer.cs:line 256
   at Nerdbank.MessagePack.MessagePackSerializer.Serialize[T](T& value, ITypeShapeProvider provider, CancellationToken cancellationToken) in D:\source\repos\Nerdbank.MessagePack\src\Nerdbank.MessagePack\MessagePackSerializer.FriendlyOverloads.cs:line 124
   at MessagePackSerializerPolyfill.Serialize[T](MessagePackSerializer serializer, T& value, CancellationToken cancellationToken) in D:\source\repos\Nerdbank.MessagePack\test\Nerdbank.MessagePack.Tests\MessagePackSerializerPolyfill.cs:line 11
   at AssemblyLoadTests.AppDomainTestDriver.SerializeSomethingSimple() in D:\source\repos\Nerdbank.MessagePack\test\Nerdbank.MessagePack.Tests\AssemblyLoadTests.cs:line 81
```

[...]

Issue #215, opened 2025-07-30
Improve witness type discoverability in netstandard2.0


eiriktsarpalis:
In netstandard2.0 compatible targets, users need to explicitly declare a witness type and look for a `ShapeProvider` property that has been inserted by the source generator. I think we can do better than that, and @m-nash's proposal of using an assembly-level attribute [pointing to the generated context](https://github.com/dotnet/runtime/issues/117379) seems to be a step in the right direction for older versions of .NET.

TL;DR the source generator would introduce an assembly-level attribute containing a reference to the `ShapeProvider` singleton like so:
```C#
[assembly: ShapeProvider(typeof(ShapeProvider), propertyName: nameof(ShapeProvider.Default))]
```
Then an AOT-friendly library could try to resolve the provider by searching the assembly containing the type, then falling back to the entry/executing assembly. For types whose shape providers are located in non-trivial locations, the existing APIs accepting IShapeProvider can be utilized.

@AArnott thoughts?

eiriktsarpalis:
> I don't see the attribute being a substitute for the StreamJsonRpc scenario that I laid out.

Rather than explicitly requiring a witness parameter, the API could try to resolve the attribute from `Assembly.GetCallingAssembly()`.

> In lieu of a predictable type name, what about being able to declare a witness class without giving it a dummy type?

I think that's doable, although we probably need a new marker attribute?

AArnott:
> the API could try to resolve the attribute from `Assembly.GetCallingAssembly()`

True, but in the Visual Studio brokered services case, that will always be wrong. The calling assembly will be the StreamJsonRpc wrapping assembly Microsoft.ServiceHub.Framework, which itself doesn't actually carry any of the contracts or shapes. Not a problem though, as that assembly will just have its own property where callers can specify the type shape provider.

I feel like we'd be doubling down on the 'close your eyes and hope it works' strategy. IIRC you were originally a bit hesitant to introduce netstandard2.0 compatible APIs because you were giving up compile-time checks for type shapes. But now if we also add overloads that divine what your type shape provider is from the calling assembly, then there's even more opportunity for runtime failure and when it fails it'll be harder for the user to understand _why_ it failed.

Again... I'm not against adding the assembly attribute. I imagine there will be scenarios where it makes sense. I'm just not sure I'd use them given the use cases of the libraries I work on.

> although we probably need a new marker attribute?

Yes, I agree.

AArnott:
[...]

PR #253, opened 2025-09-07
Improve shape discoverability in netstandard2.0

eiriktsarpalis:
Introduces a `TypeShapeProviderAttribute` that lets the source generator associate type declarations or witness types to their corresponding generated type shapes in a trim and friendly manner. This acts as a substitute for the `IShapeable<T>` interface that is not available in netstandard2.0 and enables APIs that are more similar to the modern .NET experience.

Fix https://github.com/eiriktsarpalis/PolyType/issues/215. Fix https://github.com/eiriktsarpalis/PolyType/issues/252


src/PolyType.TestCases/TestCase.cs
```diff
@@ -12,7 +12,6 @@ namespace PolyType.Tests;
 /// </summary>
 public static class TestCase
 {
-#if NET
```
eiriktsarpalis, line 15: @AArnott notice how nicely the new `ResolveDynamic` method reconciles the netstandard and modern .NET APIs here.

src/PolyType.Examples/CborSerializer/CborSerializer.cs
```diff
@@ -48,6 +48,20 @@ public static CborConverter<T> CreateConverter<T>(ITypeShapeProvider shapeProvid
     public static CborConverter<T> CreateConverterUsingReflection<T>() =>
         CreateConverter<T>(ReflectionProvider.ReflectionTypeShapeProvider.Default);
 
+    /// <summary>
+    /// Builds an <see cref="CborConverter{T}"/> instance from the specified shape.
+    /// </summary>
+    /// <typeparam name="T">The type for which to build the converter.</typeparam>
+    /// <returns>An <see cref="CborConverter{T}"/> instance.</returns>
+    /// <exception cref="NotSupportedException">No source generated implementation for <typeparamref name="T"/> was found.</exception>
+    public static CborConverter<T> CreateConverterUsingSourceGen<
```
eiriktsarpalis, line 57: @AArnott because of how generic type constraints work, this forces the introduction of separate method groups for source gen and reflection. I don't think this is necessarily a problem if designed carefully.

src/PolyType/Abstractions/TypeShapeResolver.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        return result;
+    }
+
+#if NET
+    /// <summary>
+    /// Gets the dynamic access requirements for the <see cref="ResolveDynamic{T}"/> method group.
+    /// </summary>
+    public const DynamicallyAccessedMemberTypes ResolveDynamicLinkerRequirements =
```
eiriktsarpalis, line 123: @AArnott the only drawback of the current `ResolveDynamic` implementation is that it forces all those linker requirements just to enable dynamic `IShapeable<T>` invocation. I've filed an issue about this in the [runtime repo](https://github.com/dotnet/runtime/issues/119440) for potential better alternatives, but I think we could forgo these requirements altogether if we also had the source generator emit `TypeShapeProviderAttribute` in targets that do support `IShapeable<T>`. The obvious drawback of such an approach is that it would automatically root the generated type shapes to their declarations, even if the types are never used with PolyType libraries!

PR #258, opened 2025-09-08
Optimize the SourceGenTypeShapeProvider.GetTypeShape method.

### LINKED ISSUE #252: Source generated type shape provider loads all assemblies for shaped types when only one is requested

eiriktsarpalis:
Fix #252.

### DIFF
--- a/src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.TypeShapeProvider.cs
+++ b/src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.TypeShapeProvider.cs
@@ -55,24 +55,56 @@ internal sealed partial class SourceFormatter
             /// <inheritdoc/>
             public override global::PolyType.ITypeShape? GetTypeShape(global::System.Type type)
             {
+                // This method looks up type shapes from the full transitive type graph
+                // of the current assembly, as such it can occasionally grow very large.
+                // In order to avoid performance issues with loading all types at once,
+                // group types by their names and then call a helper method that performs
+                // exact type matching with inlining disabled.
+                switch (type?.Name)
+                {
             """);
 
-        writer.Indentation++;
-
-        foreach (TypeShapeModel generatedType in provider.ProvidedTypes.Values)
+        writer.Indentation += 2;
+        foreach (var grouping in provider.ProvidedTypes.Values.GroupBy(t => t.ReflectionName).OrderBy(t => t.Key))
         {
             writer.WriteLine($$"""
-                if (type == typeof({{generatedType.Type.FullyQualifiedName}}))
+                case {{FormatStringLiteral(grouping.Key)}}:
                 {
-                    return {{generatedType.SourceIdentifier}};
+                    return MatchTypeGroup(type);
+
+                    [global::System.Runtime.CompilerServices.MethodImpl(global::System.Runtime.CompilerServices.MethodImplOptions.NoInlining)]
+                    ITypeShape? MatchTypeGroup(System.Type type)
+                    {
+                """);
+
+            writer.Indentation += 2;
+            foreach (TypeShapeModel generatedType in grouping)
+            {
+                writer.WriteLine($$"""
+                    if (type == typeof({{generatedType.Type.FullyQualifiedName}}))
+                    {
+                        return {{generatedType.SourceIdentifier}};
+                    }
+
+                    """);
+            }
+
+            writer.Indentation -= 2;
+            writer.WriteLine("""
+                        return null;
+                    }
                 }
 
                 """);
         }
 
-        writer.WriteLine("return null;");
-        writer.Indentation--;
-        writer.WriteLine('}');
+        writer.Indentation -= 2;
+        writer.WriteLine("""
+                    default:
+                        return null;
+                }
+            }
+            """);
     }
 
     private SourceText FormatGeneratedTypeMainFile(TypeDeclarationModel typeDeclaration)
--- a/src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.TypeShapeProvider.cs
+++ b/src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.TypeShapeProvider.cs
@@ -51,50 +51,37 @@ internal sealed partial class SourceFormatter
 
     private static void FormatGetShapeProviderMethod(TypeShapeProviderModel provider, SourceWriter writer)
     {
+        Debug.Assert(
+            provider.ProvidedTypes.Values.Select(t => t.ReflectionName).Distinct().Count() == provider.ProvidedTypes.Count, 
+            "The string-based type identifier should be unique to each generated type.");
+
         writer.WriteLine("""
             /// <inheritdoc/>
             public override global::PolyType.ITypeShape? GetTypeShape(global::System.Type type)
             {
-                // This method looks up type shapes from the full transitive type graph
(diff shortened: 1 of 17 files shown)


src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.TypeShapeProvider.cs
```diff
@@ -55,24 +55,56 @@ private static void FormatGetShapeProviderMethod(TypeShapeProviderModel provider
             /// <inheritdoc/>
             public override global::PolyType.ITypeShape? GetTypeShape(global::System.Type type)
             {
+                // This method looks up type shapes from the full transitive type graph
```
eiriktsarpalis: @AArnott would there be any downsides to applying the same strategy in the AssociatedTypes lookup method that current accepts string?

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.TypeShapeProvider.cs
```diff
@@ -55,24 +55,56 @@ private static void FormatGetShapeProviderMethod(TypeShapeProviderModel provider
...
 
-        writer.Indentation++;
-
-        foreach (TypeShapeModel generatedType in provider.ProvidedTypes.Values)
+        writer.Indentation += 2;
+        foreach (var grouping in provider.ProvidedTypes.Values.GroupBy(t => t.ReflectionName).OrderBy(t => t.Key))
         {
             writer.WriteLine($$"""
-                if (type == typeof({{generatedType.Type.FullyQualifiedName}}))
+                case {{FormatStringLiteral(grouping.Key)}}:
```
AArnott: Will this lead us to group `List<int>` together with `List<BigInteger>`? It seems your `ReflectionName` will only be `List``1` in that case. That will defeat your efforts to defer unwanted assembly loads in some cases. Can we do better?

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.TypeShapeProvider.cs
AArnott: Associated type lookup is very typically based on _open_ generic types, while the shapes they return are of course closed. We'll need to make sure that continues to work.

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.TypeShapeProvider.cs
eiriktsarpalis: Only if we use something akin to `Type.FullName` in the string comparison step. It's possible, but it also comes with performance ramifications.

eiriktsarpalis: I've tweaked the strategy to use the `Type.ToString()` format which seems concise enough for our purposes.

src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.cs
```diff
@@ -203,22 +203,60 @@ string GetElementTypeFQN() => valueType is { } vt
...
+    }
 
-        StringBuilder builder = new();
-        builder.Append("static associatedType => associatedType switch { ");
+    private void FormatAssociatedTypesFactory(SourceWriter writer, TypeShapeModel objectShapeModel, string factoryName)
+    {
+        Debug.Assert(objectShapeModel.AssociatedTypes.Count > 0);
+
+        writer.WriteLine($$"""
+            private global::PolyType.ITypeShape? {{factoryName}}(global::System.Type associatedType)
```
eiriktsarpalis, line 216: @AArnott heads up, I've also updated the associated shapes get method to use the same generation strategy as ITypeShapeProvider.

AArnott:
@eiriktsarpalis Were you expecting this to break anybody? It made a type shape that was previously obtainable, unobtainable:

https://github.com/AArnott/Nerdbank.MessagePack/actions/runs/17595139070/job/49985466545?pr=608#step:6:39

PR #267, opened 2025-09-11
Add support for [DataContract] annotated types.

### LINKED ISSUE #104: How would we support DataContractSerializer attributes?
Suppose we wanted to honor `[DataContract]` and `[DataMember]` etc. attributes in a serializer or any other PolyType use case. How might we go about it?

I suppose the source generator and reflection-based shape providers could be taught to honor another set of attributes. And that might be fine for something as well established as the DCS. But that approach wouldn't scale as well for other lesser attributes that we might want to support to ease migration to a PolyType-based library from a legacy one.

How hard would it be to write a reflection-based type shape provider for an arbitrary set of attributes like DCS? If it is or could be made to be relatively simple, then that might provide the migration path, and when folks wanted improved speed they could use `[GenerateShape]` attributes to use source generation instead.

This is a brainstorm -- not a high priority ask.

eiriktsarpalis:
cc @AArnott 

Fix #104.

### DIFF
--- a/docs/shape-providers.md
+++ b/docs/shape-providers.md
@@ -320,10 +320,10 @@ class PocoWithConstructors
 
 PolyType includes limited support for types using [data contract annotations](https://learn.microsoft.com/dotnet/framework/wcf/feature-details/using-data-contracts).
 
-Types annotated with `DataContractAttribute` will only include member shapes annotated with `DataMemberAttribute`, resolving property shape metadata from the relevant attribute properties where applicable.
+Types annotated with `DataContractAttribute` will only include member shapes annotated with `DataMemberAttribute`, resolving property shape metadata from the relevant attribute properties where applicable. DataContract types annotated with `KnownTypeAttribute` will be mapped to union shapes following the same rules as `DerivedTypeShapeAttribute`.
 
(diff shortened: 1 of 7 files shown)


docs/shape-providers.md
```diff
@@ -315,3 +315,13 @@ class PocoWithConstructors
     public PocoWithConstructors([ParameterShape(Name = "name")] string x1, [ParameterShape(Name = "age")] int x2);
 }
 ```
+
+### DataContract support
+
+PolyType includes limited support for types using [data contract annotations](https://learn.microsoft.com/dotnet/framework/wcf/feature-details/using-data-contracts).
+Types annotated with `DataContractAttribute` will only include member shapes annotated with `DataMemberAttribute`, resolving the relevant attribute properties where applicable.
+Correspondingly, if a type is not annotated with `DataContractAttribute` any `DataMemberAttribute` annotation will be ignored during shape resolution.
```
AArnott: Should we honor `[IgnoreMember]` on types that are _not_ attributed with `[DataContract]`? 

I think DCS basically has two patterns:

1. Attribute the type: only `[DataMember]` members are included.
2. Non-attributed type: public members included unless `[IgnoreMember]` is found on them.

eiriktsarpalis: Sure, I think that's doable.

PR #324, opened 2025-10-16
Add net472 target to PolyType

AArnott:
An explicit .NET Framework target framework is critical to effective ngen. Otherwise a .NET Framework assembly that references this one and uses mscorlib types as type arguments will have no 'home' for native code to be written in by ngen. The PolyType.dll *must* have an mscorlib assembly reference or those methods will never be ngen'd.


version.json
```diff
@@ -1,11 +1,11 @@
 {
   "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/main/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "1.0.0-rc.6",
+  "version": "1.0.0-rc.7",
```
eiriktsarpalis: Assuming we get this merged eventually, we'd want to separate the version bump into a distinct commit.

PR #345, opened 2025-11-12
Call Debugger.Launch() on failures when env var is set

AArnott:


src/PolyType.SourceGenerator/PolyTypeGenerator.cs
```diff
@@ -28,9 +32,9 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                 (node, _) => node is TypeDeclarationSyntax)
             .Collect()
             .Combine(knownSymbols)
-            .Select((tuple, token) => Parser.ParseFromGenerateShapeAttributes(tuple.Left, tuple.Right, token));
+            .Select((tuple, token) => DebugGuard(() => Parser.ParseFromGenerateShapeAttributes(tuple.Left, tuple.Right, token)));
```
eiriktsarpalis, line 35: This is allocating one closure per computation, consider having the `DebugGuard` methods accept a parameter as well.

src/PolyType.SourceGenerator/PolyTypeGenerator.cs
```diff
@@ -51,4 +55,40 @@ private void GenerateSource(SourceProductionContext context, TypeShapeProviderMo
     }
 
     public Action<TypeShapeProviderModel>? OnGeneratingSource { get; init; }
+
+    private static bool MaybeLaunchDebuggerButNeverHandleException()
+    {
+#pragma warning disable RS1035 // Do not use APIs banned for analyzers
+        if (Environment.GetEnvironmentVariable("POLYTYPE_LAUNCH_DEBUGGER_ON_EXCEPTION") is "1")
```
eiriktsarpalis, line 62: Should we be documenting these?

src/PolyType.SourceGenerator/PolyTypeGenerator.cs
AArnott: True. But on the scale of where this runs, I _really_ don't think a closure and delegate are going to show up as significant.

src/PolyType.SourceGenerator/PolyTypeGenerator.cs
AArnott: That's up to you. Externals could potentially use them to file more actionable bugs.

PR #368, opened 2025-12-06
Add support for inline and fixed arrays

### LINKED ISSUE #367: Shapes for fixed and inline arrays provide no means to access the elements
Structs with inline or fixed array members produce shapes that are unable to access the data inside them.

```cs
	[GenerateShape]
	public partial struct StructWithFixedArray
	{
		public const int Length = 10;

		public unsafe fixed byte Element0[Length];
	}

#if NET
	[GenerateShape]
	[InlineArray(Length)]
	public partial struct StructWithInlineArrayPrimitive
	{
		public const int Length = 10;

		private byte element0;
	}
#endif
```

AArnott:
Fixes #367

### DIFF
--- a/src/PolyType/ReflectionProvider/ReflectionEnumerableTypeShape.cs
+++ b/src/PolyType/ReflectionProvider/ReflectionEnumerableTypeShape.cs
@@ -361,7 +361,8 @@ internal sealed class ReflectionInlineArrayTypeShape<TArray, TElement>(int lengt
         {
             if (span.Length != length)
             {
-                throw new ArgumentException($"Expected {length} elements, but got {span.Length}.");
+                Throw(span, length);
+                static void Throw(ReadOnlySpan<TElement> span, int expectedLength) => throw new ArgumentException($"Expected {expectedLength} elements, but got {span.Length}.");
             }
 
             TArray array = default;
--- a/src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
+++ b/src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
@@ -98,9 +98,10 @@ public sealed class InlineArrayEnumerable<TArray, TElement>(TArray array, int le
         {
             get
             {
-                if (_index >= _length || _index < 0)
+                if (unchecked((uint)_index >= (uint)_length))
                 {
-                    throw new InvalidOperationException();
+                    Throw();
+                    static void Throw() => throw new InvalidOperationException();
                 }
 
                 return Unsafe.Add(ref Unsafe.As<TArray, TElement>(ref _array), _index);


src/PolyType/ReflectionProvider/ReflectionEnumerableTypeShape.cs
```diff
@@ -335,4 +339,104 @@ internal sealed class ReflectionAsyncEnumerableShape<TEnumerable, TElement>(Refl
...
+
+        public IEnumerator<TElement> GetEnumerator() => new Enumerator(array, length);
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        private sealed class Enumerator(TArray array, int length) : IEnumerator<TElement>
+        {
+            private int _index = -1;
+            private TArray _array = array;
+
+            public TElement Current => Unsafe.Add(ref Unsafe.As<TArray, TElement>(ref _array), _index);
```
eiriktsarpalis: Seems like bad things could happen if somebody were to call `Current` before `MoveNext()` or after `MoveNext()` has returned `false`. Should we be guarding for these cases?

src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Runtime.CompilerServices;
+
+namespace PolyType.SourceGenModel;
+
+/// <summary>
+/// Helper class for iterating over inline arrays.
+/// </summary>
+/// <typeparam name="TArray">The type of the inline array.</typeparam>
+/// <typeparam name="TElement">The type of the element.</typeparam>
+public sealed class InlineArrayEnumerable<TArray, TElement>(TArray array, int length) : ICollection<TElement>, IReadOnlyCollection<TElement>
```
eiriktsarpalis, line 14: This is being mostly duplicated in reflection? Why not consolidate?

src/PolyType/ReflectionProvider/ReflectionEnumerableTypeShape.cs
AArnott: Agreed. Silly Copilot.

src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
AArnott: Done.

src/PolyType/ReflectionProvider/ReflectionEnumerableTypeShape.cs
```diff
@@ -335,4 +339,42 @@ internal sealed class ReflectionAsyncEnumerableShape<TEnumerable, TElement>(Refl
     public override bool IsAsyncEnumerable => true;
     public override Func<TEnumerable, IEnumerable<TElement>> GetGetEnumerable() =>
         static _ => throw new InvalidOperationException("Sync enumeration of IAsyncEnumerable instances is not supported.");
-}
\ No newline at end of file
+}
+
+[RequiresUnreferencedCode(ReflectionTypeShapeProvider.RequiresUnreferencedCodeMessage)]
+[RequiresDynamicCode(ReflectionTypeShapeProvider.RequiresDynamicCodeMessage)]
+internal sealed class ReflectionInlineArrayTypeShape<TArray, TElement>(int length, ReflectionTypeShapeProvider provider, ReflectionTypeShapeOptions options)
```
eiriktsarpalis, line 346: Is there some kind of assert that we can use to check that `length` is the correct value for `TArray`?

src/PolyType/ReflectionProvider/ReflectionEnumerableTypeShape.cs
```diff
@@ -335,4 +339,42 @@ internal sealed class ReflectionAsyncEnumerableShape<TEnumerable, TElement>(Refl
...
+        return array => new SourceGenModel.InlineArrayEnumerable<TArray, TElement>(array, length);
+    }
+
+    public override ParameterizedCollectionConstructor<TElement, TElement, TArray> GetParameterizedConstructor()
+    {
+        return (ReadOnlySpan<TElement> span, in CollectionConstructionOptions<TElement> options) =>
+        {
+            if (span.Length != length)
+            {
+                throw new ArgumentException($"Expected {length} elements, but got {span.Length}.");
```
eiriktsarpalis: We can make the delegate body shorter by avoiding inline throws
```suggestion
                Throw();
                static void Throw() => throw new ArgumentException($"Expected {length} elements, but got {span.Length}.");
```

src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        /// <inheritdoc/>
+        public TElement Current
+        {
+            get
+            {
+                if (_index >= _length || _index < 0)
+                {
+                    throw new InvalidOperationException();
```
eiriktsarpalis: Should increase the probability of the getter being inlined by the JIT
```suggestion
                    Throw();
                    static void Throw() => throw new InvalidOperationException();
```

src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
```diff
@@ -0,0 +1,121 @@
...
+            _length = length;
+            _index = -1;
+        }
+
+        /// <inheritdoc/>
+        public TElement Current
+        {
+            get
+            {
+                if (_index >= _length || _index < 0)
```
eiriktsarpalis:
```suggestion
                if ((uint)_index >= (uint)_length)
```

src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
```diff
@@ -0,0 +1,121 @@
...
+    /// <inheritdoc/>
+    public void Add(TElement item) => throw new NotSupportedException();
+
+    /// <inheritdoc/>
+    public void Clear() => throw new NotSupportedException();
+
+    /// <inheritdoc/>
+    public bool Contains(TElement item)
+    {
+        foreach (var element in this)
```
eiriktsarpalis, line 32: I'm assuming this would lower into use of the enumerable. Should we perhaps consider a bespoke implementation?

src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
AArnott: Interesting. Copilot emitted that code initially, but I changed it thinking it was too obtuse. I'll be happy to change to that though.
It relies on C# compiling with `unchecked` operators. Is that something you want to guarantee at the project level, or should I add `unchecked( )` around the expression?

src/PolyType/ReflectionProvider/ReflectionEnumerableTypeShape.cs
AArnott: Nope. Fixed arrays have no length defined at runtime at all. Inline arrays _might_, I'm not sure. But this class is used for both.

src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
eiriktsarpalis: Good question. I always assumed unchecked was the default but it can't hurt to be proactive.

src/PolyType/ReflectionProvider/ReflectionEnumerableTypeShape.cs
AArnott: That didn't work exactly, because `static` precludes the closure that your local function expression body requires. But I fixed it.

src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
AArnott: I don't anticipate anyone will even call this method, since the contract is simply that we implement `IEnumerable<TElement>`. I only implement the collection interfaces so that `TryGetEnumeratedCount` works. So I don't think optimizing this method is worth any time unless we get feedback.

src/PolyType/SourceGenModel/Helpers/InlineArrayEnumerable.cs
AArnott: It _is_ the default, by default. But IMO it's an unsafe one. In Nerdbank.MessagePack I set `checked` to be the default in my project so that the serializer can be relied on to not silently corrupt data. In those few cases I actually _want_ `unchecked` behavior, I add the keyword to the expression.
And anytime I'm around source generators, where you don't want to bother special casing based on which setting the project is using, I always emit `checked` or `unchecked` expressions explicitly whenever it really matters.

eiriktsarpalis: Yeah, that's fair when it comes to source generated code. It's fairly standard optimization we do try to pursue in dotnet/runtime especially in hot-path getter methods.

PR #413, opened 2026-04-14
Sort ProvidedTypes iteration for deterministic source output

AArnott:
The source generator iterates `provider.ProvidedTypes.Values` in two places:

1. `SourceFormatter.AddAllSourceFiles` — emits per-type source files
2. `SourceFormatter.FormatGetShapeProviderMethod` — emits `switch` cases in `GetTypeShape`

Since `ProvidedTypes` is an `ImmutableEquatableDictionary` backed by `Dictionary<TKey, TValue>`, iteration order is non-deterministic. This causes the generated source to vary between compilations even when inputs are identical.

## Impact

This breaks incremental build for downstream projects. When a project referencing a PolyType-using library is built:

1. Any source change triggers recompilation of the library
2. PolyType source generator re-runs, producing fields/cases in a different order
3. The reference assembly content changes (different field numbering in the compiler's `<>O` display class)
4. Downstream projects see a newer reference assembly and recompile unnecessarily

In our case, a single implementation-only change in one project caused `csc` to run for a downstream project that should have been skipped entirely.

## Fix

Sort by `SourceIdentifier` (a unique, stable, type-name-derived key) before iterating in both locations. This is a two-line change.

### DIFF
--- a/src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.cs
+++ b/src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.cs
@@ -289,7 +289,7 @@ internal sealed partial class SourceFormatter(TypeShapeProviderModel provider)
             """);
 
         writer.Indentation += 2;
-        foreach (AssociatedTypeId associatedType in objectShapeModel.AssociatedTypes)
+        foreach (AssociatedTypeId associatedType in objectShapeModel.AssociatedTypes.OrderBy(t => t.ClosedTypeReflectionName))
         {
             if (associatedType.OpenTypeInfo is { } openTypeInfo)
             {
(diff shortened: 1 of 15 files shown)


src/PolyType.SourceGenerator/SourceFormatter/SourceFormatter.cs
```diff
@@ -34,7 +34,7 @@ private void AddAllSourceFiles(SourceProductionContext context, TypeShapeProvide
         context.CancellationToken.ThrowIfCancellationRequested();
         context.AddSource($"{provider.ProviderDeclaration.SourceFilenamePrefix}.g.cs", FormatTypeShapeProviderMainFile(provider));
 
-        foreach (TypeShapeModel type in provider.ProvidedTypes.Values)
+        foreach (TypeShapeModel type in provider.ProvidedTypes.Values.OrderBy(t => t.SourceIdentifier))
```
eiriktsarpalis, line 37: Do you think it might make sense to make the equatable dictionaries and sets deterministic by construction so that this doesn't regress elsewhere. The trade-off is that we'd require the key to be comparable and not just equatable.

eiriktsarpalis: Let's not block on that for now though.

AArnott:
Thanks for making this more comprehensive. :)

PR #432, opened 2026-05-18
Reduce allocations in TypeShapeResolver.ResolveDynamicOrThrow

AArnott:
The `TypeShapeResolver.ResolveDynamicOrThrow` method allocates a new type shape provider on every invocation, even though the type shape provider and type shape itself can be reasonably expected to be observably immutable.

These allocations showed up during VS startup. 

I prepared this change in three changes:

1. Cache the type shape provider instance instead of recreating it on every request for its (1) type shape. 
    *Assumption:* the type shape provider implementation will produce either an immutable type shape or a new type shape instance on every request for a type shape.
2. Shift from a _factory_ of the type shape to just caching and returning the type shape directly.
    *Assumption:* The type shape itself is immutable (and thus safe to cache and return to every caller).
3. Replace the manual caching logic with a static field initializer so the CLR can deal with the one-time initialization instead of our C# code.

@eiriktsarpalis Are the assumptions reasonable?

### DIFF
--- a/src/PolyType/Abstractions/TypeShapeResolver.cs
+++ b/src/PolyType/Abstractions/TypeShapeResolver.cs
@@ -52,7 +52,7 @@ public static class TypeShapeResolver
     [RequiresDynamicCode(ResolveDynamicMessage)]
 #endif
     public static ITypeShape<T>? ResolveDynamic<T>() =>
-        ResolveDynamicFactoryCache<T, T>.GetFactory()?.Invoke();
+        ResolveDynamicFactoryCache<T, T>.GetTypeShape();
 
     /// <summary>
     /// Uses reflection to resolve the source generated <see cref="ITypeShape{T}"/> implementation from another type.
@@ -75,7 +75,7 @@ public static class TypeShapeResolver
     [RequiresDynamicCode(ResolveDynamicMessage)]
 #endif
     public static ITypeShape<T>? ResolveDynamic<T, TProvider>() =>
-        ResolveDynamicFactoryCache<T, TProvider>.GetFactory()?.Invoke();
+        ResolveDynamicFactoryCache<T, TProvider>.GetTypeShape();
 
     /// <summary>
     /// Uses reflection to resolve the source generated <see cref="ITypeShape{T}"/> implementation of the type.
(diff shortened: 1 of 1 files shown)


src/PolyType/Abstractions/TypeShapeResolver.cs
```diff
@@ -151,30 +151,14 @@ public static ITypeShape<T> ResolveDynamicOrThrow<T, TProvider>()
 #endif
     private static class ResolveDynamicFactoryCache<T, TProvider>
     {
-        private static Func<ITypeShape<T>?>? s_cachedFactory;
```
eiriktsarpalis, line 154: The old code suggests to me I was intentionally avoiding the caching of ITypeShape instances as such, but I can't remember the exact motivations behind this. Probably not so much to anticipate potentially mutable models, but perhaps to avoid rooting unneeded allocations that come with the transitive closure of a given type and its dependencies. I think it might also be matter of not allowing such caches in the core library, but force them to happen at the library level if necessary -- this caching pattern is flexible enough in that can be implemented at any layer.

AArnott: I see. So you want to avoid caching ITypeShapeProvider in a static? I can respect that.
Are you pretty confident you don't cache them elsewhere? Maybe we should add tests to verify that these can be GC'd after calling various APIs to lock that in.
I know I've seen bugs around *types* (in collectible ALCs) being effectively anchored by static caches by various libraries, so if you're trying to avoid that, that certainly sounds valuable.
I'll see what I can do to fix this at the nerdbank.messagepack layer then.

AArnott:
Closing due to desire for PolyType to not cache things.
Replaced with https://github.com/AArnott/Nerdbank.MessagePack/pull/958

src/PolyType/Abstractions/TypeShapeResolver.cs
eiriktsarpalis: > Are you pretty confident you don't cache them elsewhere?

Not 100% confident. The only static cache in PolyType.dll I'm aware of is the reflection-based provider, but that does use ConditionalWeakTable to support unloading.

eiriktsarpalis: > I know I've seen bugs around types (in collectible ALCs) being effectively anchored by static caches by various libraries,

I would need to double check, but I suspect the fact we're caching the factory itself might be sufficient to prevent unloading.

eiriktsarpalis:
> Closing due to desire for PolyType to not cache things.

FWIW I'm still open to considering this change.

PR #454, opened 2026-06-17
Add `TypeShapeResolver.Resolve` methods for non-.NET targeting clients

AArnott:
I'm leveraging C# method overload resolution such that binding will be against the dynamic method unless the code is compiling against .NET, in which case it will bind against the preferred compile-time variants.

Projects that target .NET (perhaps in addition to .NET Framework or .NET Standard) will get an obsolete error when they use the method but are missing the attributes necessary for compile-time resolution.

At this point, `ResolveDynamicOrThrow` is more or less deprecated, as it offers less protection than `Resolve` has. In particular, `ResolveDynamicOrThrow` can be called while targeting .NET without warning, even when it's destined to fail at runtime. `Resolve` on the other hand will always be implemented in the ideal way, and if .NET is among the target frameworks of a consuming project, they'll get compile-time errors if attributes are missing that would lead to a runtime failure.
But we leave `ResolveDynamicOrThrow` without an `[Obsolete]` attribute because it is a workaround for a corner case of a library compiling that targets .NET that needs a type shape that is declared in a .NET Framework/Standard targeted assembly.


src/PolyType/Abstractions/TypeShapeExtensions.cs
```diff
@@ -18,6 +21,62 @@ public static class TypeShapeExtensions
...
+        public static ITypeShape<T> Resolve<T>()
+            => TypeShapeResolver.ResolveDynamicOrThrow<T>();
+
+        /// <inheritdoc cref="TypeShapeResolver.ResolveDynamicOrThrow{T, TProvider}()"/>
+#if NET8_0
+        [RequiresDynamicCode(TypeShapeResolver.ResolveDynamicMessage)]
+#endif
+#if NET
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("Use the RpcTargetMetadata.FromShape<T, TProvider>() method instead. If using the extension method syntax, check that your type argument actually has a [GenerateShape] attribute or otherwise implements IShapeable<T> to avoid a runtime failure.", error: true)]
```
eiriktsarpalis:
```suggestion
        [Obsolete("Use the TypeShapeProvider.Resolve<T, TProvider>() method instead. If using the extension method syntax, check that your type argument actually has a [GenerateShape] attribute or otherwise implements IShapeable<T> to avoid a runtime failure.", error: true)]
```

src/PolyType/Abstractions/TypeShapeExtensions.cs
```diff
@@ -18,6 +21,62 @@ public static class TypeShapeExtensions
...
+        /// </summary>
+        /// <typeparam name="T">The type for which a shape provider is required.</typeparam>
+        /// <typeparam name="TProvider">The type on which the shape provider is implemented.</typeparam>
+        /// <returns>An <see cref="ITypeShapeProvider"/> instance that can produce the shape that describes <typeparamref name="T"/>.</returns>
+#if NET8_0
+        [RequiresDynamicCode(TypeShapeResolver.ResolveDynamicMessage)]
+#endif
+#if NET
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("Use the RpcTargetMetadata.FromShape<T, TProvider>() method instead. If using the extension method syntax, check that your type argument actually has a [GenerateShape] attribute or otherwise implements IShapeable<T> to avoid a runtime failure.", error: true)]
```
eiriktsarpalis:
```suggestion
        [Obsolete("Use the TypeShapeProvider.ResolveProvider<T, TProvider>() method instead. If using the extension method syntax, check that your type argument actually has a [GenerateShape] attribute or otherwise implements IShapeable<T> to avoid a runtime failure.", error: true)]
```

src/PolyType/Abstractions/TypeShapeResolver.cs
```diff
@@ -146,6 +146,25 @@ public static ITypeShape<T> ResolveDynamicOrThrow<T, TProvider>()
         return result;
     }
 
+#if NET
+    /// <summary>
+    /// Resolves the <see cref="ITypeShapeProvider"/> from the <see cref="IShapeable{T}"/> implementation of another type.
+    /// </summary>
+    /// <typeparam name="T">The type for which to extract the shape provider.</typeparam>
+    /// <returns>An <see cref="ITypeShapeProvider"/> instance that can produce the shape that describes <typeparamref name="T"/>.</returns>
+    public static ITypeShapeProvider ResolveProvider<T>()
```
eiriktsarpalis: Why is this method group needed at all if you can just call `.Provider` on any shape?

src/PolyType/Abstractions/TypeShapeExtensions.cs
```diff
@@ -18,6 +21,62 @@ public static class TypeShapeExtensions
...
+        public static ITypeShapeProvider ResolveProvider<T, TProvider>()
+            => TypeShapeResolver.ResolveDynamicOrThrow<T, TProvider>().Provider;
+
+        /// <inheritdoc cref="TypeShapeResolver.ResolveDynamicOrThrow{T}()"/>
+#if NET8_0
+        [RequiresDynamicCode(TypeShapeResolver.ResolveDynamicMessage)]
+#endif
+#if NET
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("Use the TypeShapeResolver.ResolveProvider<T>() method instead. If using the extension method syntax, check that your type argument actually has a [GenerateShape] attribute or otherwise implements IShapeable<T> to avoid a runtime failure.", error: true)]
```
eiriktsarpalis: I take it that this obsolete warning is only designed to fire when invoking the extension method explicitly, and that the compiler will always resolve to the native `Resolve` method otherwise?

eiriktsarpalis:
```suggestion
        [Obsolete("Use the TypeShapeResolver.Resolve<T>() method instead. If using the extension method syntax, check that your type argument actually has a [GenerateShape] attribute or otherwise implements IShapeable<T> to avoid a runtime failure.", error: true)]
```

src/PolyType/Abstractions/TypeShapeResolver.cs
AArnott: It's not. It's slightly more direct for the StreamJsonRpc use case. But I'm happy to drop it if you think it's not worth the public API surface area.

src/PolyType/Abstractions/TypeShapeExtensions.cs
AArnott: I expect that users won't (ever) call this method directly. If they do, the Obsolete message will still steer them right.
The primary use case though is that if the user types `TypeShapeResolver.ResolveProvider<T>` and `T` is _not_ shapeable, the compiler will automatically divert to the extension method. And if targeting .NET when this steering happens, the user will get the obsolete error message so they know to add the attribute.

src/PolyType/Abstractions/TypeShapeResolver.cs
AArnott: I dropped it.

src/PolyType.Examples/JsonSerializer/JsonSerializerTSExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+    extension(JsonSerializerTS)
+    {
+        /// <summary>
+        /// Builds a <see cref="JsonConverter{T}"/> instance from the specified shape.
+        /// </summary>
+        /// <typeparam name="T">The type for which to build the converter.</typeparam>
+        /// <returns>An <see cref="JsonConverter{T}"/> instance.</returns>
+        /// <exception cref="NotSupportedException">No source generated implementation for <typeparamref name="T"/> was found.</exception>
+#if NET8_0
+        [RequiresDynamicCode("Dynamic resolution of IShapeable<T> interface may require dynamic code generation in .NET 8 Native AOT. It is recommended to switch to statically resolved IShapeable<T> APIs or upgrade your app to .NET 9 or later.")]
```
eiriktsarpalis, line 25: I'm feeling like this corner case is adding more complexity to an already complicated API matrix. Seeing as .NET 8 is going out of support soon, how do you feel about the prospect of us simply suppressing this warning in the ResolveDynamic implementation (and surfacing a relevant runtime exception message to the user?) This is a dynamic resolution method after all, so we have license to fail at runtime.

src/PolyType/Abstractions/TypeShapeExtensions.cs
```diff
@@ -18,6 +21,33 @@ public static class TypeShapeExtensions
     public static Func<object>? GetDefaultConstructor(this IObjectTypeShape shape) =>
         (Func<object>?)shape.Constructor?.Accept(DefaultConstructorVisitor.Instance);
 
+    extension(TypeShapeResolver)
```
eiriktsarpalis, line 24: I would suggest moving this extension inside `TypeShapeResolver` itself, if possible. It should serve to highlight that this is merely a workaround for the lack of .NET support for overloads with generic constraints. It would also remove the need to implement these methods in terms of methods marked obsolete.

src/PolyType.Examples/JsonSerializer/JsonSerializerTSExtensions.cs
AArnott: Without this attribute, the build breaks when targeting .NET 8. It wants to see every method that calls such decorated methods to itself be decorated.
When .NET 8 falls out of support, I'd be totally fine dropping .NET 8 as a target and removing this snippet.

src/PolyType/Abstractions/TypeShapeExtensions.cs
AArnott: Since `TypeShapeResolver` itself is `static`, that's doable. I'll make the change.

docs/CSharpSamples/Middleware.cs
```diff
@@ -0,0 +1,90 @@
...
+    internal const string ResolveDynamicMessage =
+        "Dynamic resolution of IShapeable<T> interface may require dynamic code generation in .NET 8 Native AOT. " +
+        "It is recommended to switch to statically resolved IShapeable<T> APIs or upgrade your app to .NET 9 or later.";
+
+    /// <summary>
+    /// Processes a shape of type T.
+    /// </summary>
+    /// <typeparam name="T">The type to operate on.</typeparam>
+    /// <param name="shape">The shape to process.</param>
+    public static void ConsumeShape<T>(ITypeShape<T> shape)
```
eiriktsarpalis, line 52: I would avoid recommending _public_ APIs accepting `ITypeShape` values as a best practice.

docs/CSharpSamples/Middleware.cs
```diff
@@ -0,0 +1,90 @@
...
+        public static void ConsumeShape<T>() => ShapeProcessor.ConsumeShape(TypeShapeResolver.Resolve<T>());
+
+        /// <inheritdoc cref="ShapeProcessor.ConsumeShape{T}(ITypeShape{T})"/>
+        /// <typeparam name="TProvider">The type that provides the shape for T.</typeparam>
+#if NET8_0
+        [RequiresDynamicCode(ShapeProcessor.ResolveDynamicMessage)]
+#endif
+#if NET
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("Use the ShapeProcessor.ConsumeShape<T, TProvider>() method instead. If using the extension method syntax, check that T actually has a [GenerateShape] attribute or otherwise implements or is constrained to IShapeable<T> to avoid a runtime failure.", error: true)]
```
eiriktsarpalis: If the point of this obsolete annotation is to avoid direct invocation without using extension syntax, isn't EB.Never sufficient to discourage this from happening?

docs/CSharpSamples/Middleware.cs
AArnott: Well.... that's the thing...
What I'm coming to due to the corner case you pointed out is that we need to divert users to this when T comes from a non-.NET assembly but when they themselves are .NET, so that they can switch from:

```cs
Middleware.ConsumeShape<TypeFromOldAssembly>()
```

to this:

```cs
Middleware.ConsumeShape(TypeShapeResolver.ResolveDynamic<TypeFromOldAssembly>())
```

The _only_ time they'll need this is in that one corner case. 
The alternative to this is that the middleware has to itself offer a `ConsumeShapeDynamic<T>` method, just to cater to the corner case crowd.
And FWIW, in my use cases of PolyType, it _is_ occasionally useful to have an `ITypeShape<T>` overload. Particularly when you're an RPC library like StreamJsonRpc and all you have is `ITypeShapeProvider` from which to get a shape object.

docs/CSharpSamples/Middleware.cs
AArnott: No, the point of this is to call out when C# diverts to invoking this method while targeting .NET, due to T not actually being shapeable.
The obsolete message isn't quite right in light of our recent discussions so I'm going to refine it.

src/PolyType.Examples/JsonSerializer/JsonSerializerTSExtensions.cs
eiriktsarpalis: This can be fixed by adding an `UnconditionalSuppressMessage` annotation on the ResolveDynamic method.

src/PolyType/Abstractions/TypeShapeExtensions.cs
AArnott: Actually, it's not doable. C# doesn't allow declaring two methods with the same signature within the same type, even if one is nested inside an `extension`.

docs/CSharpSamples/Middleware.cs
AArnott: And FWIW, you expose such overloads yourself in at least some of your examples.

src/PolyType/Abstractions/TypeShapeExtensions.cs
eiriktsarpalis: Can it be worked around by using an intermediate nested type?

tests/PolyType.Tests/TypeShapeResolverTests.cs
```diff
@@ -20,23 +20,23 @@ public static void ResolveDynamic_WithProvider_ReturnsExpectedSingleton()
         Assert.Same(s1, s2);
     }
 
-    [Fact]
+    [Fact, Obsolete]
```
eiriktsarpalis: What does Obsolete mean in the context of unit testing?

docs/CSharpSamples/Middleware.cs
eiriktsarpalis: > And FWIW, you expose such overloads yourself in at least some of your examples.

Right, IIRC these predate the existence of `ResolveDynamic` and should probably have been removed. As a side note, these methods act as factories that build converters so these typically wouldn't get invoked in the hot path raising questions about caching keyed on individual type shapes.

docs/CSharpSamples/Middleware.cs
eiriktsarpalis: What if the user doesn't have the option to convert because they're consuming a 3P netstandard library? This would be forcing them to regenerate using a witness type. It's a stretch of an example to be sure, but I just want to better understand the boundaries of the approach.

tests/PolyType.Tests/TypeShapeResolverTests.cs
AArnott: It suppresses warnings *within* the method from calling obsolete APIs.

src/PolyType.Examples/JsonSerializer/JsonSerializerTSExtensions.cs
AArnott: I suppose. But that's generally supposed to be used when we can expect it to *work* for other reasons that the linker can't see. In this case, I would expect it _wouldn't_ work, and we're just exchanging a compile break for a runtime exception. Why is that a good trade?

src/PolyType/Abstractions/TypeShapeExtensions.cs
AArnott: Nope. C# requires `extension` members to appear in a top-level class.

docs/CSharpSamples/Middleware.cs
AArnott: The latest iteration of the message is:

> Check that T actually has a `[GenerateShape]` attribute or otherwise implements or is constrained to `IShapeable<T>`. If T is declared in an assembly that does not target .NET, use `TypeShapeResolver.ResolveDynamic<T>()` instead.

This gives straightforward guidance either way:

1. Recompile the dependency to target .NET (preferred)
2. Or call the `ResolveDynamic<T>` method.

And when this appears in middleware, the suggested message is a bit different:

> Check that T actually has a `[GenerateShape]` attribute or otherwise implements or is constrained to `IShapeable<T>`. If T is declared in an assembly that does not target .NET, use `ShapeProcessor.ConsumeShape(TypeShapeResolver.ResolveDynamic<T>())` instead.

This unblocks them either way.

AArnott: Actually though, the equivalent alternatives would be to use `ResolveDynamicOrThrow<T>`. Otherwise `ConsumeShape` might get a `null` argument.
So maybe we shouldn't mark `ResolveDynamicOrThrow<T>` as obsolete.

docs/CSharpSamples/Middleware.cs
AArnott: Maybe. Or we could leave them, and remove the `CreateConverterUsingReflection<T>` variants. Having to have a dedicated overload called that seems a bit tacky to me. But with this more generally useful overload that takes an `ITypeShape<T>`, you not only have a reasonable workaround for the .NET Framework corner case, but you can also do the (hopefully similarly unusual) reflection type shapes with `CreateConverter(ReflectionTypeShapeProvider.Default.GetTypeShapeOrThrow<T>())`.
Or since you already have a `CreateConverter<T>(ITypeShapeProvider typeShapeProvider)` overload users can just use `CreateConverter<T>(ReflectionTypeShapeProvider.Default)`, offering you another way to get rid of `CreateConverterUsingReflection` if you wanted to cut down on method group count.

AArnott:
Closing due to lack of interest by the maintainer. Feel free to merge any time though. I personally am happy with it.

PR #456, opened 2026-06-21
Avoid representing the same required parameter twice

### LINKED ISSUE #450: Strange behaviour for constructor shape with required property
Given

```csharp
public class Base
{
	[ConstructorShape]
	public Base(string requiredString)
	{
		RequiredString = requiredString;
	}

	[PropertyShape(IsRequired= true)]
	public  string RequiredString { get; set; }

	public string OptionalString { get; set; }
}
```

Then
```
var shape = (IObjectTypeShape<Base>)PolyType.ReflectionProvider.ReflectionTypeShapeProvider.Default.GetTypeShape<Base>();

shape.Constructor.Parameters.Count is 3
```

This prevents deserializtion using *Nerdbank.MessagePack* since it will need 2 occurrences of "RequiredString"  on deserialization

If I instead change the constructor parameter to `optionalString` then it works as expected with 2 properties.

It seems to be caused by the `!memberInitializer.IsRequired`  in 
 
https://github.com/eiriktsarpalis/PolyType/blob/6b0e8defb80cebba848f071ceb2a1d7b46bb87f4/src/PolyType/ReflectionProvider/ReflectionObjectTypeShape.cs?plain=1#L208C16-L208C16

Possible solutions:
* Remove `IsRequired` from if statement (at least required from annotations) ?
* Change to `(!memberInitializer.IsRequired || setsRequiredMembers)` just as is made for properties with `required` modifier so that `SetsRequiredMembersAttribute` is a workaround

**Note**: if the member uses `required` modifider then I can be 

Is this indented behavior?

AArnott:
Consider a type has a `required` property and a constructor that takes that property as a parameter _without_ a `[SetsRequiredMembers]` attribute on the ctor:

```cs
public class SomeType
{
	public Base(string requiredString)
	{
		RequiredString = requiredString;
	}

	public required string RequiredString { get; set; }
}
```

PolyType was creating two parameter shapes: one for the parameter and one for the member initializer. Both were set as Required. As a result, middleware like Nerdbank.MessagePack failing to deserialize such types because the type shape reported not all required properties were initialized.

This change adds a test for this scenario and asserts that only one parameter shape is created, and that its `IsRequired` and `HasDefaultValue` properties have defensible values given the various test cases thrown at it.

Fixes eiriktsarpalis/PolyType#450

### DIFF
--- a/src/PolyType.Roslyn/ModelGenerator/TypeDataModelGenerator.Object.cs
+++ b/src/PolyType.Roslyn/ModelGenerator/TypeDataModelGenerator.Object.cs
@@ -359,26 +359,27 @@ public partial class TypeDataModelGenerator
                 property = property with { IsRequiredBySyntax = false };
             }
 
-            if (!property.IsRequiredBySyntax && MatchesConstructorParameter(property))
+            if (GetMatchingConstructorParameter(property) is not null && !property.IsRequiredBySyntax)
             {
-                // Deduplicate any optional properties whose signature matches a constructor parameter.
+                // Deduplicate properties whose signature matches a constructor parameter unless
+                // an object initializer slot is required to satisfy C# required members.
                 continue;
             }
 
             (memberInitializers ??= []).Add(property);
 
-            bool MatchesConstructorParameter(PropertyDataModel settableProperty)
+            IParameterSymbol? GetMatchingConstructorParameter(PropertyDataModel settableProperty)
             {
                 foreach (IParameterSymbol p in constructor.Parameters)
                 {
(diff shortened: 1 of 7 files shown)


AArnott (commented):
This PR is a draft because I haven't carefully studied the library changes made here. (Copilot authored it.) 
At the moment I'm interested in a review of the _added and changed tests_ to see if we agree on the objective.

I have confirmed that these changes make the original scenario in Nerdbank.MessagePack work.

eiriktsarpalis:
I had a fix prepared but somehow forgot to push the PR: https://github.com/eiriktsarpalis/PolyType/pull/457

PR #470, opened 2026-07-02
Fix ReflectionProvider bug on inheritance

Daniel-Svensson:
The reflection provider currently fails with 
System.InvalidOperationException is not a valid subtype of 'b' for the inheritance scenario in updated test case. The bug is present for both KnownTypeAttribute and DerivedTypeShapeAttribute that unintentionally includes attributes on base classes (in contrast to code generation provider).

**Enhancements to test coverage:**

* Added `PersianCat` as a derived type of `Cat` and updated the test data model to reflect this deeper inheritance. [[1]](diffhunk://#diff-25ba926f35c0b7abe0c934b8f20704245ba5edd8f5ff5cdd50d18acc9ec3eedaR218-R232) [[2]](diffhunk://#diff-25ba926f35c0b7abe0c934b8f20704245ba5edd8f5ff5cdd50d18acc9ec3eedaR206)
* Modified the `KnownTypeAttribute_ReportsUnionShape` test to expect three union cases (`Dog`, `Cat`, and `PersianCat`) for `Animal`.
* Added a new test `KnownTypeAttribute_DerivedTypesReportsUnionShape` to verify that only `PersianCat` is recognized as a union case for `Cat`. So that shapes from base classes are not propagated.

### DIFF
--- a/src/PolyType.SourceGenerator/Parser/Parser.cs
+++ b/src/PolyType.SourceGenerator/Parser/Parser.cs
@@ -615,6 +615,12 @@ public sealed partial class Parser : TypeDataModelGenerator
                 }
 
                 derivedType = dt;
+
+                // Only consider derived types, KnownTypeAttribute can be used for AssociatedTypes purposes
+                if (!type.IsAssignableFrom(derivedType))
+                {
+                    continue;
+                }
             }
             else
             {
--- a/src/PolyType.SourceGenerator/Parser/Parser.cs
+++ b/src/PolyType.SourceGenerator/Parser/Parser.cs
@@ -615,12 +615,6 @@ public sealed partial class Parser : TypeDataModelGenerator
                 }
 
                 derivedType = dt;
-
-                // Only consider derived types, KnownTypeAttribute can be used for AssociatedTypes purposes
-                if (!type.IsAssignableFrom(derivedType))
-                {
-                    continue;
-                }
             }
             else
             {
--- a/src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
+++ b/src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
@@ -310,8 +310,6 @@ public class ReflectionTypeShapeProvider : ITypeShapeProvider
 
                     return attr.Type;
                 })
-                // Only consider derived types, KnownTypeAttribute can be used for AssociatedTypes purposes
-                .Where(type => unionType.IsAssignableFrom(type))
                 .Select(t => new DerivedTypeShapeAttribute(t));
 
             derivedTypeAttributes.AddRange(mappedKnownTypeAttributes);
--- a/src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
+++ b/src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
@@ -294,13 +294,13 @@ public class ReflectionTypeShapeProvider : ITypeShapeProvider
             return (IUnionTypeShape)Activator.CreateInstance(fsharpUnionTypeTy, fSharpUnionInfo, this, options)!;
         }
 
-        List<DerivedTypeShapeAttribute> derivedTypeAttributes = unionType.GetCustomAttributes<DerivedTypeShapeAttribute>().ToList();
+        List<DerivedTypeShapeAttribute> derivedTypeAttributes = unionType.GetCustomAttributes<DerivedTypeShapeAttribute>(inherit: false).ToList();
 
         // Honor KnownTypeAttribute annotations only when no DerivedTypeShapeAttribute is present,
         // which takes precedence over KnownTypeAttribute.
         if (derivedTypeAttributes.Count == 0)
         {
-            var mappedKnownTypeAttributes = unionType.GetCustomAttributes<KnownTypeAttribute>()
+            var mappedKnownTypeAttributes = unionType.GetCustomAttributes<KnownTypeAttribute>(inherit: false)
                 .Select(attr =>
                 {
                     if (attr.Type is null)
@@ -308,8 +308,11 @@ public class ReflectionTypeShapeProvider : ITypeShapeProvider
                         throw new NotSupportedException("KnownTypeAttribute annotations using methods are not supported.");
                     }
 
-                    return new DerivedTypeShapeAttribute(attr.Type);
-                });
+                    return attr.Type;
+                })
+                // Only consider derived types, KnownTypeAttribute can be used for AssociatedTypes purposes
+                .Where(type => unionType.IsAssignableFrom(type) && type != unionType)
+                .Select(t => new DerivedTypeShapeAttribute(t));
 
             derivedTypeAttributes.AddRange(mappedKnownTypeAttributes);
         }
--- a/src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
+++ b/src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
@@ -311,7 +311,7 @@ public class ReflectionTypeShapeProvider : ITypeShapeProvider
                     return attr.Type;
                 })
                 // Only consider derived types, KnownTypeAttribute can be used for AssociatedTypes purposes
-                .Where(type => unionType.IsAssignableFrom(type) && type != unionType)
+                .Where(type => unionType.IsAssignableFrom(type))
                 .Select(t => new DerivedTypeShapeAttribute(t));
 
             derivedTypeAttributes.AddRange(mappedKnownTypeAttributes);
--- a/src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
+++ b/src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
@@ -308,9 +308,8 @@ public class ReflectionTypeShapeProvider : ITypeShapeProvider
                         throw new NotSupportedException("KnownTypeAttribute annotations using methods are not supported.");
                     }
 
-                    return attr.Type;
-                })
-                .Select(t => new DerivedTypeShapeAttribute(t));
+                    return new DerivedTypeShapeAttribute(attr.Type);
+                });
 
             derivedTypeAttributes.AddRange(mappedKnownTypeAttributes);
         }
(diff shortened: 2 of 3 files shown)


src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
```diff
@@ -294,22 +294,25 @@ private IUnionTypeShape CreateUnionTypeShape(Type unionType, FSharpUnionInfo? fS
...
                     {
                         throw new NotSupportedException("KnownTypeAttribute annotations using methods are not supported.");
                     }
 
-                    return new DerivedTypeShapeAttribute(attr.Type);
-                });
+                    return attr.Type;
+                })
+                // Only consider derived types, KnownTypeAttribute can be used for AssociatedTypes purposes
+                .Where(type => unionType.IsAssignableFrom(type))
```
eiriktsarpalis: This is user misconfiguration, so I would rather have this fail with an explicit error.

src/PolyType.SourceGenerator/Parser/Parser.cs
```diff
@@ -615,6 +615,12 @@ protected override IEnumerable<DerivedTypeModel> ResolveDerivedTypes(ITypeSymbol
                 }
 
                 derivedType = dt;
+
+                // Only consider derived types, KnownTypeAttribute can be used for AssociatedTypes purposes
+                if (!type.IsAssignableFrom(derivedType))
```
eiriktsarpalis: Likewise, this should probably emit a generator diagnostic.

tests/PolyType.Tests/DataContractShapeTests.cs
```diff
@@ -89,9 +89,19 @@ public void KnownTypeAttribute_ReportsUnionShape()
...
 
-        Assert.Equal(2, shape.UnionCases.Count);
+        Assert.Equal(3, shape.UnionCases.Count);
         Assert.Contains(shape.UnionCases, c => c.UnionCaseType.Type == typeof(Dog));
         Assert.Contains(shape.UnionCases, c => c.UnionCaseType.Type == typeof(Cat));
+        Assert.Contains(shape.UnionCases, c => c.UnionCaseType.Type == typeof(PersianCat));
+    }
+
+    [Fact]
+    public void KnownTypeAttribute_DerivedTypesReportsUnionShape()
```
eiriktsarpalis, line 99: How does this test for the bug you fixed?

src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
Daniel-Svensson: The code is there to allow existing classes using a not so known feature of DataContactSerializer where KnownTypeAttribute on a class actually has 2 different use cases.

A: The common use case to signal Derived types.
B: The not so common use case where it is used in the same sense as your `AssociatedTypesShapeAttribute`, it adds the types to the list of known types allowing them to be used on other properties, (such as on `object` properties)

I've seen the second usage a couple of times in a few code bases but it is not very common and I can remove it if it is important to not support the scenario.

tests/PolyType.Tests/DataContractShapeTests.cs
Daniel-Svensson: Getting the type shape for cat would throw InvalidOperationException. saying that Dog is not assignable to Cat

It would never reach the asserts.

src/PolyType.SourceGenerator/Parser/Parser.cs
Daniel-Svensson: I'll remove it.

I added it at a late addition to make in behave as reflection provider, but since the code uses GenerateTypeShape and so is already adapted to PolyType and can easily be changed, so the compat feature does not make sense.,

src/PolyType.SourceGenerator/Parser/Parser.cs
```diff
@@ -615,6 +615,12 @@ protected override IEnumerable<DerivedTypeModel> ResolveDerivedTypes(ITypeSymbol
...
+                {
+                    continue;
+                }
```
Daniel-Svensson:
```suggestion
```

tests/PolyType.Tests/DataContractShapeTests.cs
Daniel-Svensson: I did not see any similar testcases for DerivedTypeShapeAttributes, if you want separate tests to verify DerivedTypeShapeAttributes please give some advice on where they make sense to add

src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
Daniel-Svensson:
```suggestion
```

tests/PolyType.Tests/DataContractShapeTests.cs
eiriktsarpalis: Can't you use Assert.Throws?

Daniel-Svensson: @eiriktsarpalis can you please elaborate a bit,  I am new to xunit, why would I use Assert.Throws to check that a method call does not throw ? 

I verified that the test fails without the code fix, since GetTypeShape would throw an an exception.

In addition I check the union cases to match `KnownTypeAttribute_ReportsUnionShape` above since originally all cases from the base class was discovered and added, so It would guard against Cat beeing a union case.

eiriktsarpalis: Ok, I understood you had said the opposite. As mentioned in https://github.com/eiriktsarpalis/PolyType/pull/470#discussion_r3512526506 I would still prefer it if we threw an exception in this case to make sure user misconfiguration is caught immediately.

Daniel-Svensson: Ok, I think there is some confusion since the PR did 2 things.

1. The most important part is fix a problem when **it threw an exception when it should not**.

The test verifies that the following works with both KnownType (the same problem is fixed for DerivedType as well) attributes. The problem only affected ReflectionProvider but works with `GenerateShapeAttribute`.
Note that the attributes for "PersianCat" is applied to both  "Animal" and "Cat ", (which is not required by DataContractSerializer) but match what is required by json attributes and the PolyType DerivedType when used with source generator. Which is how I understand you want it to work.


```csharp
    [DataContract]
    [KnownType(typeof(Dog))]
    [KnownType(typeof(Cat))]
    [KnownType(typeof(PersianCat))]
    public partial class Animal
    {
        [DataMember(Order = 0)] public string? Name { get; set; }
        [DataMember(Order = 1)] public bool Barks { get; set; }
    }

    [GenerateShape]
    [DataContract]
    [KnownType(typeof(PersianCat))]
    public partial class Cat : Animal
    {
        [DataMember(Order = 1)] public int Lives { get; set; }
    }

    [DataContract]
    public class PersianCat : Cat
    {
        [DataMember(Order = 2)] public string? FurColor { get; set; }
        [DataMember(Order = 3)] public bool RequiresGrooming { get; set; }
    }
```


2. It stopped throwing for KnownTypeAttributes that was used in "AssociatedType" scenarios which is fully supported by DataContractSerializer. That support was removed so it will give error just as before.

src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
```diff
@@ -294,22 +294,23 @@ private IUnionTypeShape CreateUnionTypeShape(Type unionType, FSharpUnionInfo? fS
...
                     if (attr.Type is null)
                     {
                         throw new NotSupportedException("KnownTypeAttribute annotations using methods are not supported.");
                     }
 
-                    return new DerivedTypeShapeAttribute(attr.Type);
-                });
+                    return attr.Type;
+                })
+                .Select(t => new DerivedTypeShapeAttribute(t));
```
eiriktsarpalis: In its current form this is just moving part of the computation into a second select. Not needed anymore presumably?

tests/PolyType.Tests/DataContractShapeTests.cs
eiriktsarpalis: > The most important part is fix a problem when it threw an exception when it should not.

I would still like it if we threw an exception here but using a better error message. Would you like to add such a validation step?

src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
Daniel-Svensson: That is correct, the change can be reverted.

tests/PolyType.Tests/DataContractShapeTests.cs
Daniel-Svensson: @eiriktsarpalis can you please clarify what you think is wrong with the code, I don'y get why it should throw an exception.


And how you would like the code to look like to model the following classes:

```
Animal --> Cat  --> PersianCat
            |--> Dog
```


The following code works flawlessly with the PolyType Source generator, Newtonsoft.Json, System.Text.Json (change it to json equivalient attributes) and DataContractSerializer. But does throw for Reflection provider. 

In no place does the attribute placement indicate that PersianCat should be be a assignable to dog or othrewise.

```csharp
    [DataContract]
    [KnownType(typeof(Dog))]
    [KnownType(typeof(Cat))]
    public partial class Animal
    {
        [DataMember] public string? Name { get; set; }
        [DataMember] public bool Barks { get; set; }
    }

    [DataContract]
    [KnownType(typeof(PersianCat))]
    public partial class Cat : Animal
    {
      [DataMember] public int Lives { get; set; }
    }

    [DataContract]
    public class PersianCat : Cat
    {
        [DataMember] public string? FurColor { get; set; }
    }
```

eiriktsarpalis: > I don'y get why it should throw an exception.

In earlier iterations you added a check that silently filtered types if they weren't assignable to the declaring type. In other words invalid inputs such as `[KnownType(typeof(int))]` where being removed from the list. While that's an improvement over the current lack of validation, my point was that any such misconfiguration should be immediately be signalled to the user (in the case of reflection, using a clearly worded exception message explaining what is wrong).

src/PolyType/ReflectionProvider/ReflectionTypeShapeProvider.cs
eiriktsarpalis: Ok, could you make that revert please?

Daniel-Svensson: reverted

tests/PolyType.Tests/DataContractShapeTests.cs
Daniel-Svensson: thank you, I though you wanted something else since I had remove that code from this PR

